PR #4, opened 2023-10-27
Add buf-breaking-action

smallsamantha:


nicksnyder:
I think we should allow breaking changes for a bit.

.github/workflows/buf.yaml
```diff
@@ -8,12 +8,21 @@ jobs:
...
       - uses: bufbuild/buf-setup-action@v1
       - uses: bufbuild/buf-lint-action@v1.0.3
+  breaking:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: bufbuild/buf-setup-action@v1
+      - uses: bufbuild/buf-breaking-action@v1.1.2
+        with:
+          against: buf.build/bufbuild/knit
```
pkwarren, line 18:
```suggestion
          against: buf.build/bufbuild/registry
```

.github/workflows/buf.yaml
```diff
@@ -8,12 +8,21 @@ jobs:
       - uses: actions/checkout@v3
       - uses: bufbuild/buf-setup-action@v1
       - uses: bufbuild/buf-lint-action@v1.0.3
+  breaking:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: bufbuild/buf-setup-action@v1
+      - uses: bufbuild/buf-breaking-action@v1.1.2
```
pkwarren, line 16:
```suggestion
      - uses: bufbuild/buf-breaking-action@v1
```

In general I think we should update all of our action versions to `v1`.

pkwarren:
> I think we should allow breaking changes for a bit.

We could technically enable this now because `ignore_unstable_packages` set: https://github.com/bufbuild/registry-proto/blob/7775d7826dbef85929cdc8857f15fee5f0180733/buf.yaml#L13

.github/workflows/buf.yaml
```diff
@@ -8,12 +8,21 @@ jobs:
       - uses: actions/checkout@v3
       - uses: bufbuild/buf-setup-action@v1
       - uses: bufbuild/buf-lint-action@v1.0.3
+  breaking:
+    runs-on: ubuntu-20.04
```
pkwarren, line 12:
```suggestion
    runs-on: ubuntu-latest
```

We can change everything to run on latest now (which is 22.04 instead of 20.04).

.github/workflows/buf.yaml
bufdev: I'd agree

bufdev:
Conflicts need to be resolved, and we should take @pkwarren's suggestion and use major versions for GitHub Actions... @smallsamantha can you update this and then we can merge? As @pkwarren said, we can enable this since we have `ignore_unstable_packages` set to `true`.

pkwarren (approved):

PR #6, opened 2023-10-30
Add go_package

pkwarren:
In order to ensure that all generated code comes from the correct package, add the go_package option (similar to protovalidate) to all .proto files.

This should also serve as a reminder for users that generated registry code should come from generated SDKs.


rodaine (approved):

bufdev (changes requested):
We explicitly should not have to do this - this represents a failure on our part. I need to understand why this is needed, and then we should fix that problem.

bufdev (approved):

Issue #2, opened 2023-10-26
Better represent/document uniqueness of `name` on User and Organization


nicksnyder:
I was looking at CreateUsers and wondering how it was idempotent because it wasn't obvious to me that the `name` field was unique.
https://github.com/bufbuild/registry-proto/blob/601ec088649be87ee6a683602fe1e6ecc587a665/buf/registry/owner/v1beta1/user_service.proto#L82-L89

I went to look at the definition of User and still wasn't obvious the name was intended to be unique since the field documentation didn't indicate that either. I did eventually find the docs on the User type itself, but those docs are actually not very close to the field documentation:
https://github.com/bufbuild/registry-proto/blob/601ec088649be87ee6a683602fe1e6ecc587a665/buf/registry/owner/v1beta1/user.proto#L23-L42

Furthermore, I kind of expect "id" like fields to be first in the definition of a type.

Similar situation for Organization.name too.

Here are some things that we could do to improve the situation for both `User` and `Organization` (not mutually exclusive, could to some or all of them).
<details>
<summary>
Move the uniqueness documentation to the `name` field on `User` and `CreateUsersRequest.Value`
</summary>

```diff
diff --git a/buf/registry/owner/v1beta1/user.proto b/buf/registry/owner/v1beta1/user.proto
index 366b31e..77c444a 100644
--- a/buf/registry/owner/v1beta1/user.proto
+++ b/buf/registry/owner/v1beta1/user.proto
@@ -21,8 +21,6 @@ import "buf/validate/validate.proto";
 import "google/protobuf/timestamp.proto";
 
 // A user on the BSR.
-//
-// A name uniquely identifies a User, however name is mutable.
 message User {
   option (buf.registry.priv.extension.v1beta1.message).response_only = true;
 
@@ -36,6 +34,7 @@ message User {
   // The last time the User was updated.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
   // The name of the User.
+  // A name uniquely identifies a User, however name is mutable.
   string name = 4 [
     (buf.validate.field).required = true,
     (buf.validate.field).string.max_len = 255
diff --git a/buf/registry/owner/v1beta1/user_service.proto b/buf/registry/owner/v1beta1/user_service.proto
index 969f28f..a964ab1 100644
--- a/buf/registry/owner/v1beta1/user_service.proto
+++ b/buf/registry/owner/v1beta1/user_service.proto
[...]

PR #7, opened 2023-10-30
Document uniqueness of `name` on `Organization` and `User`

nicksnyder:
Document the uniqueness of the `name` field on the field itself.

Closes https://github.com/bufbuild/registry-proto/issues/2

### DIFF
--- a/buf/registry/owner/v1beta1/organization.proto
+++ b/buf/registry/owner/v1beta1/organization.proto
@@ -21,6 +21,8 @@ import "buf/validate/validate.proto";
 import "google/protobuf/timestamp.proto";
 
 // Organization is an organization on the BSR.
+//
+// A name uniquely identifies an Organization, however name is mutable.
 message Organization {
   option (buf.registry.priv.extension.v1beta1.message).response_only = true;
 
--- a/buf/registry/owner/v1beta1/organization.proto
+++ b/buf/registry/owner/v1beta1/organization.proto
@@ -29,16 +29,16 @@ message Organization {
     (buf.validate.field).required = true,
     (buf.validate.field).string.uuid = true
   ];
+  // The time the Organization was created.
+  google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
+  // The last time the Organization was updated.
+  google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
   // The name of the Organization.
   // A name uniquely identifies an Organization, however name is mutable.
-  string name = 2 [
+  string name = 4 [
     (buf.validate.field).required = true,
     (buf.validate.field).string.max_len = 255
   ];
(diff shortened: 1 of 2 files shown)


bufdev:
I could go either way on the meat of this - moving the unique fields above the `create_time` and `update_time`. The thinking was that the "standard resource fields", namely `id, create_time, update_time`, all come first, where the resource-specific fields come later.

If we did do this, you'd need to update the comments globally, not just for one package (ie all of the `module` package needs to be updated as well). However, I do think having the uniqueness constraints at the top level on a message is nice, so we could perhaps have them both on the field and at the top level.

nicksnyder:
> The thinking was that the "standard resource fields", namely id, create_time, update_time, all come first, where the resource-specific fields come later.

Yeah, I figured.

> If we did do this, you'd need to update the comments globally, not just for one package (ie all of the module package needs to be updated as well). 

Not sure what you mean here. It isn't clear to me that we would want to re-order `name` field in `branch`, `module`, `tag` because those are only identifiers within some context (user/org/module), and not really equivalent to an `id` or `name` in the same sense that it is for `User` and `Organization`.

In any case, after some reflection, I think adding docs to the field is sufficient (which is what this PR diff now is), and we don't need to go so far as changing the order of or renaming the fields.

bufdev:
> Not sure what you mean here.

Yea looking into it, we actually do have "uniqueness" comments on the `module` package types, I wasn't paying attention closely enough. This LGTM.

buf/registry/owner/v1beta1/organization.proto
```diff
@@ -36,6 +36,7 @@ message Organization {
   // The last time the Organization was updated.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
   // The name of the Organization.
+  // A name uniquely identifies an Organization, however name is mutable.
```
bufdev, line 39: Can you put a newline between 38 and 38, with a comment? Same below. So:

```
// The name of the Organization.
//
// A name uniquely identifies an Organization, however name is mutable.
```

nicksnyder: Yep

```suggestion
  //
  // A name uniquely identifies an Organization, however name is mutable.
```

buf/registry/owner/v1beta1/user.proto
```diff
@@ -36,6 +36,7 @@ message User {
   // The last time the User was updated.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
   // The name of the User.
+  // A name uniquely identifies a User, however name is mutable.
```
nicksnyder, line 39:
```suggestion
  //
  // A name uniquely identifies a User, however name is mutable.
```

bufdev (approved):

Issue #1, opened 2023-10-26
Document transactional nature of Create* RPCs


nicksnyder:
Batch create requests like [CreateUsersRequest](https://github.com/bufbuild/registry-proto/blob/601ec088649be87ee6a683602fe1e6ecc587a665/buf/registry/owner/v1beta1/user_service.proto#L82C1-L82C27) are intended to have a transactional nature where they either all success or none of them do. It would be nice to document these semantics for batch requests like this in the proto file.

PR #11, opened 2023-10-30
Document atomic nature of batch create/update/delete RPCs

nicksnyder:
Closes https://github.com/bufbuild/registry-proto/issues/1

### DIFF
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -181,6 +181,8 @@ message CreateCommitsRequest {
   //
   // If there are already VCSCommits on the associated Module with a given hash, this
   // will result in an error. Otherwise, a new VCSCommit is created.
+  //
+  // This operation is atomic. Either all Commits are created, or an error happens and none of them are created.
   repeated AssociatedVCSCommit associated_vcs_commits = 5;
 }
 
--- a/buf/registry/module/v1beta1/module_service.proto
+++ b/buf/registry/module/v1beta1/module_service.proto
@@ -112,6 +112,8 @@ message CreateModulesRequest {
     string release_branch_name = 6 [(buf.validate.field).string.max_len = 255];
   }
   // The Modules to create.
+  //
+  // This operation is atomic. Either all Modules are created, or an error happens and none of them are created.
   repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
 }
 
@@ -148,6 +150,8 @@ message UpdateModulesRequest {
     }
   }
   // The Modules to update.
+  //
+  // This operation is atomic. Either all Modules are updated, or an error happens and none of them are updated.
   repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
 }
 
@@ -158,6 +162,8 @@ message UpdateModulesResponse {
 
 message DeleteModulesRequest {
   // The Modules to delete.
+  //
+  // This operation is atomic. Either all Modules are deleted, or an error happens and none of them are deleted.
   repeated ModuleRef module_refs = 1 [(buf.validate.field).repeated.min_items = 1];
 }
 
--- a/buf/registry/module/v1beta1/tag_service.proto
+++ b/buf/registry/module/v1beta1/tag_service.proto
@@ -101,6 +101,8 @@ message CreateTagsRequest {
     ];
   }
   // The Tags to create.
+  //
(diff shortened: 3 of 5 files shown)


buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -181,6 +181,8 @@ message CreateCommitsRequest {
   //
   // If there are already VCSCommits on the associated Module with a given hash, this
   // will result in an error. Otherwise, a new VCSCommit is created.
+  //
+  // This operation is atomic. Either all Commits are created, or an error happens and none of them are created.
```
bufdev, line 185: I think this comment is on the wrong spot - this comment is on the associated VCS commit field, I think you mean to say that all Commits in the workspace, as WELL AS all associated VCS commits, are created atomically, and this comment should likely be on the RPC or Request level.

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -35,10 +35,13 @@ service CommitService {
   rpc ListCommitHistory(ListCommitHistoryRequest) returns (ListCommitHistoryResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
-  // Create commits on a Module with associated Content.
+  // Create commits on a Module with associated content (e.g. blobs, branches, tags).
   //
   // This is used by push and sync.
   //
+  // This operation is atomic. Either all Commits and associated content are created, or an error happens and none of
+  // them are created.
```
nicksnyder, line 43: @saquibmian @unmultimedio @doriable We need to document the atomicity of this request. In this PR I am presuming that this is atomic, but want to verify with you that doing this atomically is both possible and desirable. Can you confirm?

buf/registry/module/v1beta1/commit_service.proto
nicksnyder: Good catch. Since RPCs already have doc strings and the request objects don't, I moved the docs to the RPC for this one. It does beg the question on whether the other comments should get moved to the RPC level as well.

buf/registry/module/v1beta1/commit_service.proto
saquibmian: Yes, and yes.

nicksnyder: Thanks!

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -35,10 +35,13 @@ service CommitService {
   rpc ListCommitHistory(ListCommitHistoryRequest) returns (ListCommitHistoryResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
-  // Create commits on a Module with associated Content.
+  // Create commits on a Module with associated content (e.g. blobs, branches, tags).
```
bufdev, line 38: branches and tags aren't content - can you just revert this?

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -35,10 +35,13 @@ service CommitService {
...
   //
   // This is used by push and sync.
   //
+  // This operation is atomic. Either all Commits and associated content are created, or an error happens and none of
```
bufdev, line 42: `or an error is returned.`

buf/registry/module/v1beta1/commit_service.proto
nicksnyder:
```suggestion
  // Create commits on a Module with associated Content.
```

buf/registry/module/v1beta1/commit_service.proto
nicksnyder:
```suggestion
  // This operation is atomic. Either all Commits and associated content are created, or an error is returned.
```

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -39,6 +39,9 @@ service CommitService {
   //
   // This is used by push and sync.
   //
+  // This operation is atomic. Either all Commits and associated content are created, or an error is returned.
+  // them are created.
```
nicksnyder, line 43:
```suggestion
```

buf/registry/module/v1beta1/commit_service.proto
nicksnyder: I moved all docs to be on the RPC method since that seems to make more sense and is more consistent.

bufdev (approved):

PR #15, opened 2023-11-03
Update buf to version with protovalidate linter

pkwarren:


Makefile
```diff
@@ -10,7 +10,7 @@ BIN := .tmp/bin
 export PATH := $(BIN):$(PATH)
 export GOBIN := $(abspath $(BIN))
 
-BUF_VERSION := v1.27.1
+BUF_VERSION := 21ba19590be4afe6fa3af0406e138c4fdc36fedf
```
pkwarren, line 13: Will change to a new version when the new CLI release is available.

rodaine (approved):

jchadwick-buf (approved):
Seems OK to me. The actual validation rules will need to be updated still to match whatever range of length we actually want. (e.g. minimum for owner name may be >1) but that's not fully hashed out atm.

bufdev (approved):

PR #23, opened 2023-11-06
Validation for paths

jchadwick-buf:
* Rejects paths with backslashes
* Limits length to 4096 (will be enforced in v1alpha1 as well.)
* Enforces relative paths (no preceding slash.)
* Enforces no trailing slash
* Enforces normalized paths (no . or .. elements)

## Explanation
Regular expression breakdown:
```
Path := PathSegment(/PathSegment)*
PathSegment := [^/.][^/]?|[^/][^/.]|[^/]{3,}
```

* `Path` enforces normalized, relative paths by ensuring that there is one or more path segments with one and only one `/` between them and nowhere else.
* `PathSegment` enforces each path segment can't be `.` or `..` and must be non-empty. It does this by matching any 1 character segment that isn't `.`, any 2 character segment that isn't `..`, and any 3 character segment.

Overall path length is then enforced with `max_len` and Windows-style paths are rejected using `not_contains` to generally disallow backslashes in paths.

### DIFF
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -252,7 +252,7 @@ message GetCommitNodesRequest {
     // The separator "/" must be used.
     repeated string paths = 2 [(buf.validate.field).repeated.items = {
       string: {
-        max_len: 1000
+        max_len: 4096
         not_contains: "\\"
         pattern: "^([^/.][^/]?|[^/][^/.]|[^/]{3,})(/([^/.][^/]?|[^/][^/.]|[^/]{3,}))*$"
       }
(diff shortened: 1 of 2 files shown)


buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -250,10 +250,12 @@ message GetCommitNodesRequest {
     //
     // The path must be relative, and connect contain any "." or ".." components
     // The separator "/" must be used.
-    //
-    // TODO: Can we encode the above requirements via validation?
     repeated string paths = 2 [(buf.validate.field).repeated.items = {
-      string: {max_len: 4096}
+      string: {
+        max_len: 1000
```
bufdev, line 255: Why 1000? I think this is a breaking change to the Buf CLI, and we chose 4096 as it had to do with Windows (I believe 4096 was the max file length).

jchadwick-buf: 4096 SGTM. Not the max on Win32 (complicated) but the max in Git and both are far larger than any path we've seen yet.

bufdev (approved):

PR #24, opened 2023-11-07
Remove UserServerRole

mfridman:
This PR removes `UserServerRole` in its entirety from the API surface. Putting this up for review/feedback, although need to think about this.

I'd go as far as removing the "special" handling of empty refs in the 3 endpoints and adding limits like other endpoints.

- ListModulesRequest
- ListOrganizationsRequest
- ListUsersRequest

Having the `UserServerRole` be a core primitive in the API is a potential footgun because in various places we're going to have conditional checks to expose, hide, or mask this field. Which may leak users with elevated privileges.

It feels there should be a dedicated `AdminService` either in the registry API or even better, an internal set of protos that aren't widely publicized. It'll be less error-prone to secure a whole service or RPC than an individual field.

### DIFF
--- a/buf/registry/owner/v1beta1/user.proto
+++ b/buf/registry/owner/v1beta1/user.proto
@@ -54,20 +54,15 @@ message User {
     (buf.validate.field).required = true,
     (buf.validate.field).enum.defined_only = true
   ];
-  // The role that the User has at the BSR instance level.
-  UserServerRole server_role = 7 [
-    (buf.validate.field).required = true,
-    (buf.validate.field).enum.defined_only = true
-  ];
   // The configurable description of the User.
-  string description = 8 [(buf.validate.field).string.max_len = 350];
+  string description = 7 [(buf.validate.field).string.max_len = 350];
   // The configurable URL that represents the homepage for a User.
-  string url = 9 [
+  string url = 8 [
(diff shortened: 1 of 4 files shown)


buf/registry/owner/v1beta1/user.proto
```diff
@@ -54,20 +54,15 @@ message User {
     (buf.validate.field).required = true,
     (buf.validate.field).enum.defined_only = true
   ];
-  // The role that the User has at the BSR instance level.
-  UserServerRole server_role = 7 [
```
mfridman, line 58: Say we do keep this, should this be:

- hidden: display unspecified to all users, except if you're logged in and are a server admin
- masked: always display "member", unless logged in and are a server admin
- always show correct server role if they are authenticated
 
And the last option is,

- always show the server role, regardless if authenticated or anonymous

bufdev: If we can get away with not needing it for now, I think you're right that that is optimal.

bufdev (approved):
Assuming you want to do this, I'm fine with removing at least for now.

mfridman:
Ye, let's remove this for now and revisit it before publishing a stable v1. Always easy to add fields.

PR #25, opened 2023-11-07
Adjust validation for module name length

jchadwick-buf:
This matches the newly-unified v1alpha1 limits.

`required` does not seem to be necessary when `min_len` is present.

### DIFF
--- a/buf/registry/module/v1beta1/module.proto
+++ b/buf/registry/module/v1beta1/module.proto
@@ -38,12 +38,10 @@ message Module {
   // The name of the Module.
   //
   // Unique within a given User or Organization.
-  string name = 4 [
-    (buf.validate.field).string = {
-      min_len: 2,
-      max_len: 100
-    }
-  ];
+  string name = 4 [(buf.validate.field).string = {
+    min_len: 2,
+    max_len: 100
+  }];
   // The id of the User or Organization that owns the Module.
   string owner_id = 5 [
     (buf.validate.field).required = true,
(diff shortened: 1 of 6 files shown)


bufdev:
PR title says this is just for module names, but this seems to be for all resource names - is this what we want @mfridman ?

buf/registry/module/v1beta1/module.proto
```diff
@@ -38,10 +38,10 @@ message Module {
   // The name of the Module.
   //
   // Unique within a given User or Organization.
-  string name = 4 [
-    (buf.validate.field).required = true,
-    (buf.validate.field).string.max_len = 255
-  ];
+  string name = 4 [(buf.validate.field).string = {
+    min_len: 2,
+    max_len: 100
```
mfridman, line 43: Are we sure we want to increase the user or organization name to 100? E.g., GH limits this to 39 chars iirc.

I think module makes sense at 100.

jchadwick-buf:
@bufdev 

> PR title says this is just for module names, but this seems to be for all resource names - is this what we want @mfridman ?

I did not intend to impact any resource names that are not module names, could you please point me to which one isn't? I may have gotten confused here.

mfridman:
This looks reasonable, we're enforcing that a module name is 2-100 in places where that reference is used.

bufdev:
Nevermind, I didn't look carefully enough. LGTM.

bufdev (approved):

mfridman (approved):

PR #27, opened 2023-11-07
Make Digest a first-class citizen

bufdev:
@saquibmian context:

Buckle up, here we go!

### A `Digest` uniquely identifies the content it closes over

First, a `Digest` uniquely identifies the content it closes over. Here's how we get to this conclusion:

1. Every `Commit` has a set of `Digests`.
    - For `Digest`s of type shake256 -- which is the `Digest` we write in lock files today -- this includes:
        - all protos (file and path)
        - `buf.yaml` (or other config file)
        - `buf.lock`
        - the single documentation file path
        - license file and path
    - For `Digests` of type b3, this includes:
        - all protos 
        - dependencies (module identity and commit name)
        - the single documentation file and path
        - license file and path
        - breaking and lint configs
    - We will refer to the content that the digest closes over as `Content`.
2. Two `Commit`s with identical `Content` (i.e., config, files, and dependencies) will produce the same `Digest`. Today we use this property to short-circuit creating a new `Commit` if nothing in your module has changed.
2. Therefore, a `Digest` uniquely identifies a `Content`, i.e., the content it closes over.

### A `Digest` is a resolvable reference

1. Given a particular `Digest`, we can resolve a set of `Commit`s whose set of `Digest`s contain that value. The commits will have varying _metadata_ (repository, create time, branch, author, etc.), but the `Content` of the commits is identical -- and in fact even hashes to the same location in the CAS store.
    - In other words, a `Commit` is really just a pairing of some metadata and `Content`, and `Content` is reused across commits (sometimes, not all the time) and may even exist in various registries (this has implications for federation).
2. Given that a `Digest` uniquely identifies the content it closes over, and that a `Commit` is a pairing of some metadata and the content it closes over, we can resolve the set of `Commit`s associated with a `Digest`.
3. If there is only one `Commit` with that `Digest`, we have resolved a `Digest` to a `Commit`.
4. If there is more than one `Commit` with that `Digest`, we can further resolve to a _single_ `Commit` by considering only the _latest Released_ `Commit`.
5. If we do this, this makes all `Digests` resolvable references. Importantly, this also makes the `Digest` a resolvable _dependency_ reference.

### Tieing this all together

The fundamental problem with Workspace Push is that `Modules` have to be pushed in topological order, and if a new `Commit` is created, the `buf.lock` file for the consuming `Module` has to be updated to pin to that `Commit` before it can be pushed.

We can Push a Workspace if we can do all of this in a deterministic fashion, client-side. Today, because a `ModulePin` has the `Commit`'s `name`, we can't. We have needed the name to resolve the `ModulePin` to a `Commit`, but we have just proven that we can resolve the `ModulePin` to a `Commit` via a `Digest` instead.

Therefore, if we drop `Commit.name` from a `ModulePin`, we can determinstically generate a `ModulePin` to a `Commit` _that hasn't even been pushed yet_. All that remains is to Push in topological order, which is the easy part.

To implement this, we do the following:

1. Add `Digest` to `ResourceRef.Name`. If a `Digest` is specified, the name is interpreted to reference a `Commit` with that `Digest`.
2. Update all `List*` requests to include valid semantics for when a `Digest` is passed as a reference:
    - for `Branch`s, include all `Branch`s that contain a `Commit` with that `Digest`.
    - for `Tag`s, include all `Tag`s that are associated with `Commit`s with that `Digest`.
    - for `VCSCommit`, include all `VSCCommit`s that are associated with `Commit`s with that `Digest`.
    - for `Commit`, include the latest released `Commit` that has that `Digest`.
3. Update `CreateCommits` to do the following:
    1. Change `DepNode`, replacing `Commit.ID` with `Digest`. All dependencies references `Digest`s.
    2. Move `DepNode`s to be part of a `ModuleNode`. A `ModuleNode` would include a `DepNode` for all declared dependencies (pruned or not, doesn't matter) _and a `DepNode` for any workspace modules it imports, even though they haven't been pushed_.

With this change, as long as `ModuleNode`s are sent in topological order, and the BSR respects that ordering, multiple commits can be created in a single transaction and resolve dependencies (by `Digest`) that were also created in that transaction, thereby enabling Workspace Push.

This only works with `buf.yaml` v2, because Workspace dependencies are not in the `buf.lock`, but this can _also_ work with `buf.yaml` v1 if we choose to allow `buf mod update` on a Workspace, and it would update `buf.lock` files to what they _should_ be post Push. As long as we push in topological order, one at a time, this will all work.

Let me know if any of this is crazy, in my mind all of this works, but it needs to be thoroughly vetted to make sure it's sane.

### DIFF
--- a/buf/registry/module/v1beta1/branch_service.proto
+++ b/buf/registry/module/v1beta1/branch_service.proto
@@ -84,6 +84,7 @@ message ListBranchesRequest {
   //   - If a Tag is referenced, all Branches that contain the Tag are returned.
   //   - If a VCSCommit is referenced, all Branches that contain the VCSCommit are returned.
   //   - Is a Branch is referenced, this Branch is returned.
+  //   - If a Digest is referenced, all Branches that contain a Commit with this Digest are returned.
   ResourceRef resource_ref = 3 [(buf.validate.field).required = true];
 }
 
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -78,6 +78,8 @@ message ResolveCommitsRequest {
   //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
   //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
   //   - Is a Branch is referenced, this is interpreted to mean the latest Commit on the Branch.
+  //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
+  //     Digests referencing unreleased Commits cannot be referenced.
   repeated ResourceRef resource_refs = 1 [
     (buf.validate.field).repeated.min_items = 1,
     (buf.validate.field).repeated.max_items = 250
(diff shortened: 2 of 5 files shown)


saquibmian:
@bufdev Update the comment, take a look and let me know.

PR #31, opened 2023-11-09
Fix missing blobs request/response

saquibmian:
Blobs are not globally addressable across the BSR, largely for SOC2 compliance so that we can safely and cheaply delete all assets for a repository. Blobs are scoped to a repository.

Therefore, when we query for missing blobs, we have to query for missing blobs in the context of a `ModuleRef`. I chose to take the route of querying for multiple sets of module blobs in a single request rather than one request per set of module blobs.

### DIFF
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -310,7 +310,8 @@ message GetBlobsResponse {
 }
 
 message GetMissingBlobDigestsRequest {
-  message ModuleBlobDigests {
+  // A set of blob Digests associated with a Module.
+  message Value {
     // The reference to the Module that the digests belong to.
     ModuleRef module_ref = 1 [(buf.validate.field).required = true];
     // The digests to see if we have Blobs for.
@@ -319,19 +320,20 @@ message GetMissingBlobDigestsRequest {
       (buf.validate.field).repeated.max_items = 250
     ];
   }
-  repeated ModuleBlobDigests module_blob_digests = 1 [
+  repeated Value values = 1 [
     (buf.validate.field).repeated.min_items = 1,
     (buf.validate.field).repeated.max_items = 250
   ];
 }
 
 message GetMissingBlobDigestsResponse {
-  message MissingBlobDigest {
+  // A set of missing blob Digests associated with a Module.
+  message Value {
     // The Module that the missing digests are for.
     Module module = 1 [(buf.validate.field).required = true];
     // The digests that are missing.
-    repeated buf.registry.storage.v1beta1.Digest missing_blob_digests = 2;
+    repeated buf.registry.storage.v1beta1.Digest missing_blob_digests = 2 [(buf.validate.field).repeated.min_items = 1];
   }
   // The digests that are missing.
-  repeated MissingBlobDigest missing_blob_digests = 1;
+  repeated Value values = 1;
 }
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -319,7 +319,10 @@ message GetMissingBlobDigestsRequest {
       (buf.validate.field).repeated.max_items = 250
     ];
   }
-  repeated ModuleBlobDigests module_blob_digests = 1;
+  repeated ModuleBlobDigests module_blob_digests = 1 [
+    (buf.validate.field).repeated.min_items = 1,
+    (buf.validate.field).repeated.max_items = 250
+  ];
 }
 
 message GetMissingBlobDigestsResponse {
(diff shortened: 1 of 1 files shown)


buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -310,14 +310,25 @@ message GetBlobsResponse {
...
+  message ModuleBlobDigests {
+    // The reference to the Module that the digests belong to.
+    ModuleRef module = 1;
+    // The digests to see if we have Blobs for.
+    repeated buf.registry.storage.v1beta1.Digest blob_digests = 2 [
+      (buf.validate.field).repeated.min_items = 1,
+      (buf.validate.field).repeated.max_items = 250
+    ];
+  }
+  repeated ModuleBlobDigests module_blob_digests = 1;
```
saquibmian, line 322: Let me add `repeated` validation rules here.

unmultimedio (approved):

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -310,14 +310,28 @@ message GetBlobsResponse {
 }
 
 message GetMissingBlobDigestsRequest {
-  // The digests to see if we have Blobs for.
-  repeated buf.registry.storage.v1beta1.Digest digests = 1 [
+  message ModuleBlobDigests {
+    // The reference to the Module that the digests belong to.
+    ModuleRef module = 1;
```
bufdev, line 315: Should be `module_ref`, matching the rest of the API surface area.

When editing the APIs, be careful to make sure you are being consistent at all times.

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -310,14 +310,28 @@ message GetBlobsResponse {
...
+  repeated ModuleBlobDigests module_blob_digests = 1 [
     (buf.validate.field).repeated.min_items = 1,
     (buf.validate.field).repeated.max_items = 250
   ];
 }
 
 message GetMissingBlobDigestsResponse {
+  message MissingBlobDigest {
+    // The reference to the Module that the missing digests are for.
+    ModuleRef module = 1;
```
saquibmian, line 331: @bufdev what type would I use here? `ModuleRef` is marked a `request_only`.

Is this canonical?
```proto
  string module_id = 6 [
  ];
```
If so, that's incongruent with `ModuleRef`, which may not be an ID. Ideally the caller doesn't have to make other calls to fetch IDs for the modules (if that's the case, why take in `ModuleRef` at all).

buf/registry/module/v1beta1/commit_service.proto
saquibmian: Fixed

buf/registry/module/v1beta1/commit_service.proto
bufdev: It should be `Module module = 1` on the response. Good catch.

bufdev: and required

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -310,14 +310,28 @@ message GetBlobsResponse {
...
+  repeated ModuleBlobDigests module_blob_digests = 1 [
     (buf.validate.field).repeated.min_items = 1,
     (buf.validate.field).repeated.max_items = 250
   ];
 }
 
 message GetMissingBlobDigestsResponse {
+  message MissingBlobDigest {
+    // The Module that the missing digests are for.
+    Module module = 1 [(buf.validate.field).required = true];
```
unmultimedio, line 331: Quite a lot of info just to say "is this the module you requested?", but ok.

unmultimedio (approved):

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -310,14 +310,28 @@ message GetBlobsResponse {
...
+    ];
+  }
+  repeated ModuleBlobDigests module_blob_digests = 1 [
     (buf.validate.field).repeated.min_items = 1,
     (buf.validate.field).repeated.max_items = 250
   ];
 }
 
 message GetMissingBlobDigestsResponse {
+  message MissingBlobDigest {
```
bufdev, line 329: This isn't a single digest, this is a set of Digests, so this name doesn't make as much sense. I would want to call it `ModuleDigests`, but then the field `repeated ModuleDigests module_digests` doesn't really make sense, it's more like `repeated ModuleDigests module_digests_es`. Plurals for message names rarely make sense.

Perhaps something like `ModuleBlobDigestSet`, and then `repeated ModuleBlobDigestSet module_blob_digest_sets`, or just `Value` and `repeated Value values`.

saquibmian: Done

bufdev (approved):

unmultimedio (approved):

bufdev:
`GetBlobs` also needed fixing, this is now done.

PR #32, opened 2023-11-14
Allow tags to move

saquibmian:
Sync requires this behaviour. I also added the `UpdateTags` RPC.

We maintain the inability to reference a Tag on an unreleased Commit, which is consistent with v1alpha1.

### DIFF
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -216,8 +216,8 @@ message CreateCommitsRequest {
   }];
   // The names of Tags that should be associated with this Commit.
   //
-  // If a Tag currently exists on the associated Module with a name, the Tag will move to this
-  // commit. If the Tag does not currently exist, a new Tag will be created for the name.
+  // If a Tag currently exists on the associated Module with a name, the Tag will be updated
+  // to reference this commit. Otherwise, a new Tag will be created for the name.
   repeated string tag_names = 4 [(buf.validate.field).repeated.items = {
     string: {max_len: 250}
   }];
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -78,11 +78,11 @@ message ResolveCommitsRequest {
   //   - If a Module is referenced, this is interpreted to mean the latest released Commit on the Module.
   //   - If a Commit is referenced, this just references that specific commit.
   //   - If a Tag is referenced, this is interpreted to mean the released Commit associated with the Tag.
-  //     Tags referencing unreleased commits cannot be referenced.
+  //     Tags referencing unreleased commits cannot be resolved.
   //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
   //   - Is a Branch is referenced, this is interpreted to mean the latest Commit on the Branch.
   //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
-  //     Digests referencing unreleased Commits cannot be referenced.
+  //     Digests referencing unreleased Commits cannot be resolved.
   repeated ResourceRef resource_refs = 1 [
     (buf.validate.field).repeated.min_items = 1,
     (buf.validate.field).repeated.max_items = 250
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -77,7 +77,8 @@ message ResolveCommitsRequest {
   // Once the resource is resolved, this reference is interpeted to refer to the following Commit:
   //   - If a Module is referenced, this is interpreted to mean the latest released Commit on the Module.
   //   - If a Commit is referenced, this just references that specific commit.
-  //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
+  //   - If a Tag is referenced, this is interpreted to mean the released Commit associated with the Tag.
+  //     Tags referencing unreleased commits cannot be referenced.
   //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
   //   - Is a Branch is referenced, this is interpreted to mean the latest Commit on the Branch.
   //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
@@ -215,9 +216,8 @@ message CreateCommitsRequest {
   }];
   // The names of Tags that should be associated with this Commit.
   //
-  // If a Tag currently exists on the assocated Module with a name, the RPC will error, however
-  // this will change in the future when we allow Tags to move. If the Tag does not
-  // currently exist, a new Tag will be created for each name.
+  // If a Tag currently exists on the associated Module with a name, the Tag will move to this
+  // commit. If the Tag does not currently exist, a new Tag will be created for the name.
   repeated string tag_names = 4 [(buf.validate.field).repeated.items = {
     string: {max_len: 250}
   }];
--- a/buf/registry/module/v1beta1/commit_service.proto
+++ b/buf/registry/module/v1beta1/commit_service.proto
@@ -77,8 +77,7 @@ message ResolveCommitsRequest {
   // Once the resource is resolved, this reference is interpeted to refer to the following Commit:
   //   - If a Module is referenced, this is interpreted to mean the latest released Commit on the Module.
   //   - If a Commit is referenced, this just references that specific commit.
-  //   - If a Tag is referenced, this is interpreted to mean the released Commit associated with the Tag.
-  //     Tags referencing unreleased commits cannot be resolved.
+  //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
   //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
   //   - Is a Branch is referenced, this is interpreted to mean the latest Commit on the Branch.
   //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
--- a/buf/registry/module/v1beta1/tag_service.proto
+++ b/buf/registry/module/v1beta1/tag_service.proto
@@ -39,10 +39,10 @@ service TagService {
   rpc CreateTags(CreateTagsRequest) returns (CreateTagsResponse) {
     option idempotency_level = IDEMPOTENT;
   }
-  // Create new Tags on a Module, and Update existing Tags on a Module.
+  // Update existing Tags on a Module.
   //
-  // This operation is atomic. Either all Tags are created/updated or an error is returned.
-  rpc CreateOrUpdateTags(CreateOrUpdateTagsRequest) returns (CreateOrUpdateTagsResponse) {
+  // This operation is atomic. Either all Tags are updated or an error is returned.
+  rpc UpdateTags(UpdateTagsRequest) returns (UpdateTagsResponse) {
     option idempotency_level = IDEMPOTENT;
   }
   // Delete existing Tags.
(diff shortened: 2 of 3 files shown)


buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -120,6 +126,34 @@ message CreateTagsResponse {
...
+  message Value {
+    // The Module associated with the Tag.
+    ModuleRef module_ref = 1 [(buf.validate.field).required = true];
+    // The Tag name.
+    string name = 2 [
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.max_len = 250
+    ];
+    // The id of the Commit to associate with the Tag.
+    string commit_id = 3 [
```
unmultimedio, line 140: Instead of moduleref+commit, whay about a single resourceref like proposed in #33? Same suggestion for the Create RPC.

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -120,6 +126,34 @@ message CreateTagsResponse {
   repeated Tag tags = 1 [(buf.validate.field).repeated.min_items = 1];
 }
 
+message MoveTagsRequest {
+  // An individual request to move a Tag.
+  message Value {
+    // The Module associated with the Tag.
+    ModuleRef module_ref = 1 [(buf.validate.field).required = true];
+    // The Tag name.
+    string name = 2 [
```
unmultimedio, line 135:
```suggestion
    repeated string names = 2 [
```

We will send multiple tags for the same resourceref/commitref. Also, could you update this as well in the Create RPC?

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -39,6 +39,12 @@ service TagService {
   rpc CreateTags(CreateTagsRequest) returns (CreateTagsResponse) {
     option idempotency_level = IDEMPOTENT;
   }
+  // Move existing Tags on a Module.
+  //
+  // This operation is atomic. Either all Tags are moved or an error is returned.
+  rpc MoveTags(MoveTagsRequest) returns (MoveTagsResponse) {
```
unmultimedio, line 45:
```suggestion
  rpc CreateOrUpdateTags(CreateOrUpdateTagsRequest) returns (CreateOrUpdateTagsResponse) {
```

buf/registry/module/v1beta1/tag_service.proto
saquibmian: Yup just did it :)

buf/registry/module/v1beta1/tag_service.proto
saquibmian: Yup, just did :)

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -95,18 +101,26 @@ message ListTagsResponse {
...
   // An individual request to create a Tag.
   message Value {
-    // The Module to create the Tag under.
-    ModuleRef module_ref = 1 [(buf.validate.field).required = true];
     // The Tag name.
-    string name = 2 [
+    //
+    // If there is already a Tag with a name, the request errors. Otherwise, a new Tag will be
+    // created associated with the commit identified.
+    repeated string name = 1 [
```
unmultimedio, line 108:
```suggestion
    repeated string names = 1 [
```

and field docs

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -120,6 +134,43 @@ message CreateTagsResponse {
...
 
+message CreateOrUpdateTagsRequest {
+  // An individual request to create or update a Tag.
+  message Value {
+    // The Tag names.
+    //
+    // If there is already a Tag with a name, it will be updated to be associated with
+    // the commit identified. Otherwise, a new Tag will be created associated with the
+    // commit identified.
+    repeated string name = 1 [
```
unmultimedio, line 145:
```suggestion
    repeated string names = 1 [
```

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -101,18 +101,26 @@ message ListTagsResponse {
 message CreateTagsRequest {
   // An individual request to create a Tag.
   message Value {
-    // The Module to create the Tag under.
-    ModuleRef module_ref = 1 [(buf.validate.field).required = true];
     // The Tag name.
-    string name = 2 [
+    //
+    // If there is already a Tag with a name, the request errors. Otherwise, a new Tag will be
```
saquibmian, line 106: Ah this makes this no longer idempotent. @bufdev thoughts on this? It's also hard to make this idempotent now with the `ResourceRef` in play, because the resource may have changed.

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -39,6 +39,12 @@ service TagService {
   rpc CreateTags(CreateTagsRequest) returns (CreateTagsResponse) {
     option idempotency_level = IDEMPOTENT;
   }
+  // Create new Tags on a Module, and Update existing Tags on a Module.
+  //
+  // This operation is atomic. Either all Tags are created/updated or an error is returned.
+  rpc CreateOrUpdateTags(CreateOrUpdateTagsRequest) returns (CreateOrUpdateTagsResponse) {
```
bufdev, line 45: My question from [#33](https://github.com/bufbuild/registry-proto/pull/33#discussion_r1393337177) remains: is there any real downside to the non-atomic `Create` and then `Update` approach?

bufdev:
CI isn't passing as well.

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -95,18 +101,26 @@ message ListTagsResponse {
...
+    // See the documentation on Ref for resource resolution details.
+    //
+    // The Tags are set depending on the kind of resource that is resolved:
+    //   - If a Module is referenced, Tags are set to the latest released Commit in the module.
+    //   - If a Commit is referenced, Tags are set for that Commit.
+    //   - If a Tag is referenced, Tags are set for the Commit that this Tag references.
+    //   - If a VCSCommit is referenced, Tags are set for the Commit that this VCSCommit references.
+    //   - Is a Branch is referenced, Tags are set for the latest Commit on the Branch.
+    //   - If a Digest is referenced, Tags are set for the most recent Commit with this Digest.
+    ResourceRef resource_ref = 2 [(buf.validate.field).required = true];
```
bufdev, line 123: Why do we need to do this? Why can't I get the commit first?

buf/registry/module/v1beta1/tag_service.proto
unmultimedio: The downside here is that `Update` should only be used when the Tag already exist, so from the client perspective you'd need to `Get` the tag first to know if you should invoke `Create` or `Update`.

buf/registry/module/v1beta1/tag_service.proto
unmultimedio: You could resolve the commit id from the VCS ref, yes, but that's one more query the client could save if they pass the `ResourceRef` directly here... although as @saquibmian mentions, that takes away the idempotent property because the ref could change across requests.

saquibmian: Chatted with Peter, I'll update this.

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -77,11 +77,12 @@ message ResolveCommitsRequest {
   // Once the resource is resolved, this reference is interpeted to refer to the following Commit:
   //   - If a Module is referenced, this is interpreted to mean the latest released Commit on the Module.
   //   - If a Commit is referenced, this just references that specific commit.
-  //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
+  //   - If a Tag is referenced, this is interpreted to mean the released Commit associated with the Tag.
+  //     Tags referencing unreleased commits cannot be resolved.
   //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
   //   - Is a Branch is referenced, this is interpreted to mean the latest Commit on the Branch.
   //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
-  //     Digests referencing unreleased Commits cannot be referenced.
```
bufdev, line 84: This was correct before

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -77,11 +77,12 @@ message ResolveCommitsRequest {
   // Once the resource is resolved, this reference is interpeted to refer to the following Commit:
   //   - If a Module is referenced, this is interpreted to mean the latest released Commit on the Module.
   //   - If a Commit is referenced, this just references that specific commit.
-  //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
+  //   - If a Tag is referenced, this is interpreted to mean the released Commit associated with the Tag.
```
bufdev, line 80: Why released?

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -77,11 +77,12 @@ message ResolveCommitsRequest {
...
+  //     Tags referencing unreleased commits cannot be resolved.
```
bufdev, line 81: Why?

buf/registry/module/v1beta1/commit_service.proto
saquibmian: This reads weird, though.
> Digests referencing unreleased Commits cannot be referenced.

... but I just referenced it. It _can_ reference an unreleased commit, but you cannot _resolve_ an unreleased commit.

buf/registry/module/v1beta1/commit_service.proto
saquibmian: Same reason as digests, we don't want to allow pinning to something that isn't released.

I wonder if this should be a CLI concern, though (for this case and for the Digest case below).

bufdev: That's a caller implementation detail unless I'm missing something. Should not affect the API.

saquibmian: Sure, so just to be clear, I would update both comments, for Tags and for Digests, right?

bufdev: What we choose to allow pinning to in our business logic/implementetion has no implications on this API endpoint. The API does not take any positions on what it chosen to be implemented on top of it. If your specific implementation needs to rely on a Commit being released, it can check the release status of a Commit.

If I reference a Tag, I get back the Commit for that Tag, released or not. We have defined Digest (if I remember) to only mean released commits, although perhaps we should revisit if that restriction is necessary.

saquibmian: Agreed that whether a commit is released can be a caller concern. Aside from how tricky this is going to get in some cases (i.e., how do I ask for the latest released commit on a branch), I am trying to be consistent and am finding these rules inconsistent:

> If a Module is referenced, this is interpreted to mean the latest released Commit on the Module.

Okay, so this is released commits only.

> If a Commit is referenced, this just references that specific commit.

Okay, I guess this is fair, but this begs the question why you're trying to resolve a Commit. I find this part of the API very strange.

> If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.

Okay, so this is not released commit.

> Is a Branch is referenced, this is interpreted to mean the latest Commit on the Branch.

Assuming we go ahead with the new buf sync mental model, is this going to be constrained to a released commit?

> If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.

Okay, so this is released commits only.

Why are we sometimes constraining to released commits and sometimes not?

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -120,6 +126,26 @@ message CreateTagsResponse {
...
 }
 
+message UpdateTagsRequest {
+  // An individual request to update a Tag.
+  message Value {
+    // The Tag to update.
+    TagRef tag_ref = 1 [(buf.validate.field).required = true];
+    // The id of the Commit associated with the Tag.
+    optional string commit_id = 2 [(buf.validate.field).string.uuid = true];
+  }
```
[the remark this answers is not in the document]
saquibmian: Yes

nicksnyder:
Closing this in favor of https://github.com/bufbuild/registry-proto/pull/50

PR #33, opened 2023-11-14
Replace `CreateTags` with `SetTags`

unmultimedio:
An alternative to https://github.com/bufbuild/registry-proto/pull/32.

For tags most of the times we don't need two different RPCs, and for sync we always override/move the current tag position, or just create it if it doesn't exist.

In the case we ever need to create-only or move-only, the client can use the existing `GetTag` RPC first.

### DIFF
--- a/buf/registry/module/v1beta1/tag_service.proto
+++ b/buf/registry/module/v1beta1/tag_service.proto
@@ -16,7 +16,6 @@ syntax = "proto3";
 
 package buf.registry.module.v1beta1;
 
-import "buf/registry/module/v1beta1/module.proto";
 import "buf/registry/module/v1beta1/resource.proto";
 import "buf/registry/module/v1beta1/tag.proto";
 import "buf/validate/validate.proto";
@@ -100,11 +99,11 @@ message SetTagsRequest {
     // See the documentation on Ref for resource resolution details.
     //
     // The Tags are set depending on the kind of resource that is resolved:
-    //   - If a Module is referenced, Tags are set to the latest released commit in the module.
+    //   - If a Module is referenced, Tags are set to the latest released Commit in the module.
     //   - If a Commit is referenced, Tags are set for that Commit.
     //   - If a Tag is referenced, Tags are set for the Commit that this Tag references.
     //   - If a VCSCommit is referenced, Tags are set for the Commit that this VCSCommit references.
-    //   - Is a Branch is referenced, Tags are set for the Commit that this Branch references.
+    //   - Is a Branch is referenced, Tags are set for the latest Commit on the Branch.
     //   - If a Digest is referenced, Tags are set for the most recent Commit with this Digest.
     ResourceRef resource_ref = 1 [(buf.validate.field).required = true];
     // The Tags names.
--- a/buf/registry/module/v1beta1/tag_service.proto
+++ b/buf/registry/module/v1beta1/tag_service.proto
@@ -33,10 +33,10 @@ service TagService {
   rpc ListTags(ListTagsRequest) returns (ListTagsResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
-  // Create new Tags on a Module.
+  // Set Tags on a ResourceRef. If tags already existed, it moves them.
   //
-  // This operation is atomic. Either all Tags are created or an error is returned.
-  rpc CreateTags(CreateTagsRequest) returns (CreateTagsResponse) {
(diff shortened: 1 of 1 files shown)


buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -92,31 +92,36 @@ message ListTagsResponse {
...
+    // The reference to resolve the Commit to which this Tags will be set.
+    //
+    // See the documentation on Ref for resource resolution details.
+    //
+    // The Tags are set depending on the kind of resource that is resolved:
+    //   - If a Module is referenced, Tags are set to the latest released commit in the module.
+    //   - If a Commit is referenced, Tags are set for that Commit.
+    //   - If a Tag is referenced, Tags are set for the Commit that this Tag references.
+    //   - If a VCSCommit is referenced, Tags are set for the Commit that this VCSCommit references.
+    //   - Is a Branch is referenced, Tags are set for the Commit that this Branch references.
```
saquibmian, line 107:
```suggestion
    //   - Is a Branch is referenced, Tags are set for the latest Commit on the Branch this references.
```

saquibmian (commented):
I like the use of `ResourceRef`, even though it's a little strange when used with a `Tag` reference. 

The only caveat with `MoveTags` is that it's impossible to atomically perform a `Create` operation.

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -92,31 +91,36 @@ message ListTagsResponse {
...
+    //   - If a VCSCommit is referenced, Tags are set for the Commit that this VCSCommit references.
+    //   - Is a Branch is referenced, Tags are set for the latest Commit on the Branch.
+    //   - If a Digest is referenced, Tags are set for the most recent Commit with this Digest.
+    ResourceRef resource_ref = 1 [(buf.validate.field).required = true];
+    // The Tags names.
+    repeated string name = 2 [
       (buf.validate.field).required = true,
       (buf.validate.field).string.max_len = 250
     ];
-    // The id of the Commit associated with the Tag.
```
bufdev, line 105: How can you create a commit without a commit_id to reference? Every Tag has a commit id.

buf/registry/module/v1beta1/tag_service.proto
```diff
@@ -92,31 +91,36 @@ message ListTagsResponse {
   repeated Tag tags = 2;
 }
 
-message CreateTagsRequest {
-  // An individual request to create a Tag.
+message SetTagsRequest {
```
bufdev, line 94: Why not just `UpdateTags`, in the same style as other RPCs?

buf/registry/module/v1beta1/tag_service.proto
unmultimedio: Because the commit would be resolved by the passed `ResourceRef`. This was motivated mainly by Buf Sync in which we have `map[git.Hash][]tags`.

buf/registry/module/v1beta1/tag_service.proto
unmultimedio: Because semantically `Update` infers that the Tag existed before, I think? This RPC creates or moves the tag.

bufdev: `CreateOrUpdateTags` then

saquibmian: I think the idea behind this was that we don't care if it exists or not, just put it here. Without this, the flow is
1. `GetTags`
2. If Tag doesn't exist, call `CreateTag`; otherwise call `UpdateTags`.

I'm fine with either approach, especially because sync will _already_ call `GetTags` (to only sync tags that have changed).

bufdev: Are there any real downsides to the non-atomic approach?

saquibmian: If we have:
- `Create` create a tag that previously didn't exist
- `CreateOrUpdate` creates any tags that previously didn't exist, updates any that previously existed

Then I'm happy. I think `Create` needs to stay because it's impossible to say "create this tag and I want to be sure I'm not moving the tag and potentially breaking someone".

unmultimedio: I'm cool either way. I'll close this one in favor of #32 then, and comment there.

unmultimedio:
Closing in favor of #32 to keep atomic version: `Create` RPC.

PR #34, opened 2023-11-14
Adjust page token validation

jchadwick-buf:
We wanted to either pick some arbitrary large value or remove it entirely. Others pointed out that 255 is short enough that it might be a bit too small for base64'd cursor pagination, so I picked 4096 as an arbitrary value.

The justification for this is that _removing_ validation later would be backwards compatible, but adding validation later would be backwards incompatible. So a lenient limit seems like a decent idea.

I think this is the last of the validation issues. I made a couple additional passes to make sure and didn't see anything else that obviously needed attention.

### DIFF
--- a/buf/registry/module/v1beta1/branch_service.proto
+++ b/buf/registry/module/v1beta1/branch_service.proto
@@ -73,7 +73,7 @@ message ListBranchesRequest {
   // The page to start from.
   //
   // If empty, the first page is returned,
-  string page_token = 2;
+  string page_token = 2 [(buf.validate.field).string.max_len = 4096];
   // The reference to list Branches for.
   //
   // See the documentation on Ref for resource resolution details.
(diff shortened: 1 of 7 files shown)


bufdev:
> The justification for this is that removing it later would be backwards compatible, but adding it later would not be. So a lenient limit seems like a decent idea.

I think it's the opposite, right?

jchadwick-buf:
If you want it removed, I'm happy to go that way, too; nobody really had much of a consensus. But I think it would be a breaking change to add validation: it would reject requests that used to be considered valid. Right? Whereas removing validation shouldn't break any existing clients.

bufdev (approved):
I was just saying I think you got the wording backwards in the PR description, I think we're on the same page but you have the inverse in the description :-)

jchadwick-buf:
Ah, alright. I tried to make the PR description a bit clearer just for posterity's sake.

PR #35, opened 2023-11-15
Add ignore_empty for optional uri field validation

emcfarlane:
Adds to all uri fields that can be optional set. Using with options still adds ignore_empty to allow unsetting the value.

Otherwise empty string "" will fail validation with: 
```
invalid_argument: validation error:
 - values[0].url: value must be a valid URI [string.uri]
```

### DIFF
--- a/buf/registry/module/v1beta1/module_service.proto
+++ b/buf/registry/module/v1beta1/module_service.proto
@@ -113,7 +113,8 @@ message CreateModulesRequest {
     // The configurable URL in the description of the module.
     string url = 5 [
       (buf.validate.field).string.uri = true,
-      (buf.validate.field).string.max_len = 255
+      (buf.validate.field).string.max_len = 255,
+      (buf.validate.field).ignore_empty = true
     ];
     // The name of the release Branch of the module.
     //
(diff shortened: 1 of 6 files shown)


buf/registry/owner/v1beta1/user_service.proto
```diff
@@ -107,7 +107,8 @@ message CreateUsersRequest {
     // The configurable URL that represents the homepage for a User.
     string url = 4 [
       (buf.validate.field).string.uri = true,
-      (buf.validate.field).string.max_len = 255
+      (buf.validate.field).string.max_len = 255,
+      (buf.validate.field).ignore_empty = true
```
mfridman, line 111: Curious, why does this have to be empty? There is no `min_len` set on the field?

jchadwick-buf: Because otherwise the uri validator fails: it does not accept an empty value as a valid URI.

jchadwick-buf (approved):
Tested running protovalidate on the user create RPC and user update RPC. This seems to do what is intended.

buf/registry/owner/v1beta1/user_service.proto
mfridman: Got it! That makes sense.

emcfarlane: An empty string isn't a valid URI so fails the string on string uri: `values[0].url: value must be a valid URI [string.uri]`.

buf/registry/module/v1beta1/module_service.proto
```diff
@@ -146,7 +147,8 @@ message UpdateModulesRequest {
     // uThe configurable URL in the description of the module.
```
mfridman, line 147: Typo (unrelated to PR).

emcfarlane: Nice catch!

bufdev (approved):

PR #39, opened 2023-11-24
Add ordering for list methods

emcfarlane:
Allow consumers to sort list methods with a new `order` field on request messages. `Order` is an enum of either ascending or descending.


buf/registry/owner/v1beta1/user_service.proto
```diff
@@ -82,6 +82,23 @@ message ListUsersRequest {
...
+  Filter filter = 4;
+  enum OrderDirection {
+    ORDER_DIRECTION_UNSPECIFIED = 0;
+    ORDER_DIRECTION_ASC = 1;
+    ORDER_DIRECTION_DESC = 2;
+  }
+  // The order to sort the Users.
+  //
+  // If empty, the default ORDER_DIRECTION_ASC is used.
+  OrderDirection order_direction = 5 [(buf.validate.field).enum.defined_only = true];
```
emcfarlane, line 101: `order_direction` over `order` to be able to potentially add `order_by` later if needed. Maybe `order` is simpler.

buf/registry/owner/v1beta1/user_service.proto
```diff
@@ -82,6 +82,23 @@ message ListUsersRequest {
...
   repeated OrganizationRef organization_refs = 3;
+  message Filter {
+    // Filter by the type of the User.
+    repeated UserType type = 1 [(buf.validate.field).enum.defined_only = true];
+    // Filter by the state of the User.
+    repeated UserState state = 2 [(buf.validate.field).enum.defined_only = true];
+  }
+  // The filter to apply.
+  Filter filter = 4;
+  enum OrderDirection {
```
bufdev, line 93: Why would we add this specifically for the ListUsers endpoint?

This is incongruent with the rest of the registry API - if we want to make such a change, propose something/a pattern that universally applies

bufdev (changes requested):

buf/registry/owner/v1beta1/user_service.proto
emcfarlane: Created a shared enum `Order` in a new package `buf.registry.v1beta1` to apply the enum on each list request message.

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -86,6 +87,10 @@ message ListBranchesRequest {
   //   - Is a Branch is referenced, this Branch is returned.
   //   - If a Digest is referenced, all Branches that contain a Commit with this Digest are returned.
   ResourceRef resource_ref = 3 [(buf.validate.field).required = true];
+  // The order to sort the Branches.
+  //
+  // If empty, the default ORDER_ASCENDING is used.
+  buf.registry.v1beta1.Order order = 4 [(buf.validate.field).enum.defined_only = true];
```
bufdev, line 93: Just make all of these `bool sort_desc` - there's only two orders for sorting, there isn' the possibility of a third, we want to default to ascending, might as well just make it a `bool`, and if for some reason this doesn't work out, we can later deprecate this field

bufdev (approved):

PR #42, opened 2023-11-30
Extract module full name into its own message

unmultimedio:
When referencing module resources the module owner and name are always present, as well in the module ref.


saquibmian (commented):
I could be wrong, but I think we want to keep `*Ref.Name` fully self-contained as a consistent pattern. One indication of this is that all `*Ref`s today are `request_only`, and perhaps there's some concern that mixing them with other types gets inconsistent as the API evolves? I'll defer to @bufdev on this one.

PR #43, opened 2023-11-30
Revamp CreateCommits to just create Commits and add specialized Push and Rebase endpoints

saquibmian:
There are two things going on in this PR:

1. `CreateCommits` is updated so that it simply creates Commits for Modules without pushing them onto any Branch or assigning any Tags to them. 
   - The idea is that Commits can be "uploaded" to the BSR before they are placed in their correct places in Branch history or assigned Tags. This is very useful for `sync`, where potentially 1000s of Commits need to be created before pushing them on a Branch. Because Branch history should be patched atomically and with a lock, creating Commits in-band becomes untenable.
   - These Commits can only be referenced by their `ID`, `Digest`, or a `VCSCommitRef` if there was any `AssociatedVCSCommit`.
2. Two new RPCs are added to mutate a `Branch`'s history: `BranchService.Push` and `BranchService.Rebase`.
     - `Push` provides no assertions on the final state of a `Branch`'s history. This is the model that `buf push` currently uses, where all concurrent writes win, but the last write is king and becomes the latest commit.
     - `Rebase` allows the caller to assert on the final state of a `Branch`'s history. Concurrent writes are idempotent as long as they assert the same final state. If they don't, the first write wins and all others lose. This is the model that `buf sync` will use.

### DIFF
--- a/buf/registry/module/v1beta1/branch.proto
+++ b/buf/registry/module/v1beta1/branch.proto
@@ -60,12 +60,12 @@ message Branch {
   // Only one Branch per module will be marked as the release Branch.
   // TODO: enum?
   bool is_release = 9 [(buf.validate.field).required = true];
-  // The id of the latest Commit created on the Branch.
+  // The id of the head Commit created on the Branch.
   string head_commit_id = 10 [
     (buf.validate.field).required = true,
     (buf.validate.field).string.uuid = true
   ];
-  // The Digest of the latest Commit pushed to the Branch.
+  // The Digest of the head Commit pushed to the Branch.
   buf.registry.storage.v1beta1.Digest head_commit_digest = 11 [(buf.validate.field).required = true];
 }
 
--- a/buf/registry/module/v1beta1/branch.proto
+++ b/buf/registry/module/v1beta1/branch.proto
@@ -61,12 +61,12 @@ message Branch {
   // TODO: enum?
   bool is_release = 9 [(buf.validate.field).required = true];
   // The id of the latest Commit created on the Branch.
-  string latest_commit_id = 10 [
+  string head_commit_id = 10 [
     (buf.validate.field).required = true,
     (buf.validate.field).string.uuid = true
   ];
   // The Digest of the latest Commit pushed to the Branch.
-  buf.registry.storage.v1beta1.Digest latest_commit_digest = 11 [(buf.validate.field).required = true];
+  buf.registry.storage.v1beta1.Digest head_commit_digest = 11 [(buf.validate.field).required = true];
 }
 
 // BranchRef is a reference to a Branch, either an id or a fully-qualified name.
--- a/buf/registry/module/v1beta1/branch_service.proto
+++ b/buf/registry/module/v1beta1/branch_service.proto
@@ -17,8 +17,10 @@ syntax = "proto3";
 package buf.registry.module.v1beta1;
 
 import "buf/registry/module/v1beta1/branch.proto";
+import "buf/registry/module/v1beta1/commit.proto";
 import "buf/registry/module/v1beta1/module.proto";
 import "buf/registry/module/v1beta1/resource.proto";
+import "buf/registry/storage/v1beta1/storage.proto";
 import "buf/validate/validate.proto";
 
 option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/module/v1beta1";
@@ -37,6 +39,16 @@ service BranchService {
   rpc ListBranches(ListBranchesRequest) returns (ListBranchesResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // Push existing Commits onto Branches.
+  //
+  // This operation is atomic. Either all Commits are pushed to the referenced Branches or an error
+  // is returned.
+  rpc PushCommits(PushCommitsRequest) returns (PushCommitsResponse);
+  // Create or update histories for Branches.
+  //
+  // This operation is atomic. Either all Branches have their histories created/update updated or an
+  // error is returned.
+  rpc CreateOrUpdateBranchHistories(CreateOrUpdateBranchHistoriesRequest) returns (CreateOrUpdateBranchHistoriesResponse);
 }
 
 message GetBranchesRequest {
@@ -97,4 +109,108 @@ message ListBranchesResponse {
   string next_page_token = 1 [(buf.validate.field).string.max_len = 4096];
   // The listed Branches.
   repeated Branch branches = 2;
+}
+
+message PushCommitsRequest {
+  // A request to push a Commit to a Branch.
+  message Value {
+    // The Branch to push the Commit to.
+    BranchRef branch_ref = 1 [(buf.validate.field).required = true];
+    // The ID of the Commit to push to the Branch.
+    //
+    // This Commit will become the latest Commit on the Branch.
+    //
+    // If the Commit already exists on the Branch, an error is returned.
+    string commit_id = 2 [
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.uuid = true
+    ];
+  }
+  // The requests to push a Commit to a Branch.
+  //
+  // Commits will be pushed to Branches in the order provided.
+  repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
+}
+
+message PushCommitsResponse {
+  message Value {
+    // The Branch to which the Commit was pushed.
+    Branch branch = 1 [(buf.validate.field).required = true];
+    // The Commit that was pushed to the Branch.
+    Commit commit = 2 [(buf.validate.field).required = true];
+  }
+  // The Branches that were pushed to, in the order of the request.
+  repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
+}
+
+message CreateOrUpdateBranchHistoriesRequest {
+  // A request to creatqe a new Branch with the expected history.
+  message CreateBranchOp {
+    // The module for which to create the Branch.
+    ModuleRef module_ref = 1 [(buf.validate.field).required = true];
+    // The name of the branch to create.
+    string branch_name = 2 [
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.max_len = 250
+    ];
+    // The ID of another Commit on another Branch from which to import Branch history. The history
+    // of any Branch that contains this Commit will be imported into the newly created Branch, up to
+    // and including this Commit.
+    //
+    // This is primarily designed for use in forking a Branch from an existing Branch instead of
+    // pushing the new Branch's entire history.
+    optional string import_from_commit_id = 3 [(buf.validate.field).string.uuid = true];
+    // The IDs of Commits to add to the Branch, ordered in order in which they should be added to
+    // the Branch's history.
+    repeated string commit_ids_to_add = 4 [(buf.validate.field).string.uuid = true];
+  }
+
+  // A request to update an existing Branch's history.
+  message UpdateBranchOp {
+    // The branch to update the history of.
+    BranchRef branch_ref = 1 [(buf.validate.field).required = true];
+    // The ID of the latest Commit on the Branch.
+    //
+    // If the ID provided does not match the ID of the latest Commit on the Branch, an error is
+    // returned.
+    string latest_commit_id = 2 [
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.uuid = true
+    ];
+    // The IDs of Commits to remove from the Branch, ordered in order in which they appear in the
+    // Branch's history.
+    //
+    // These commits must appear in order at the end of the Branch's history.
+    repeated string commit_ids_to_remove = 3 [(buf.validate.field).string.uuid = true];
+    // The IDs of Commits to add to the Branch, ordered in order in which they should be added to
+    // the Branch's history.
+    repeated string commit_ids_to_add = 4 [(buf.validate.field).string.uuid = true];
+  }
+
+  message Value {
+    oneof request {
+      // A request to create a new Branch with the expected history.
+      CreateBranchOp new_branch_op = 1;
+      // A request to patch the history of a Branch.
+      //
+      // After this request is successful, the history of the Branch will be:
+      // [..., head, <commits to add>].
+      UpdateBranchOp existing_branch_op = 2;
+    }
+  }
+  // The requests to update Branch histories.
+  //
+  // Branches will be created/updated in the order provided.
+  repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
+}
+
+message CreateOrUpdateBranchHistoriesResponse {
+  message Value {
+    // The Branch whose history was updated.
+    Branch branch = 1 [(buf.validate.field).required = true];
+    // The new Commits after the previous HEAD, in the order that the Commits appear in history.
+    repeated Commit commits = 3;
+  }
+  // The Branches that were updated, in the order of the request.
+  repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
 }
--- a/buf/registry/module/v1beta1/branch_service.proto
+++ b/buf/registry/module/v1beta1/branch_service.proto
@@ -127,7 +127,7 @@ message PushRequest {
     //
     // The last Commit will become the head Commit on the Branch.
     //
-    // If the Commit already exists on the Branch, an error is returned.
+    // If any of these Commits already exist on the Branch, an error is returned.
     repeated string commit_ids = 2 [
       (buf.validate.field).repeated.min_items = 1,
       (buf.validate.field).repeated.items = {
--- a/buf/registry/module/v1beta1/branch_service.proto
+++ b/buf/registry/module/v1beta1/branch_service.proto
@@ -51,7 +51,7 @@ service BranchService {
   //
   // This operation is atomic. Either all Branches are rebased as represented in the given Operations,
   // or an error is returned. Additionally, each update in the request asserts the expected head Commit
-  // of the branch. If the actual head Commit does not match the expected head COmmit, all Operations
+  // of the branch. If the actual head Commit does not match the expected head Commit, all Operations
   // are rejected and an error is returned.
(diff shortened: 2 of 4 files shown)


buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -37,6 +39,24 @@ service BranchService {
...
+  //
+  // This operation is atomic. Either all Commits are pushed to the referenced Branches in the order
+  // provided or an error is returned.
+  //
+  // This operation does not allow the caller to assert the state of Branches before pushing any
+  // Commits onto them. As a result, multiple concurrent PushCommits operations that target the same
+  // Branch will all succeed, and the latest Commit on the Branch will be determined by the last
+  // request that succeeded. To deterministically push Commits onto a Branch, use
+  // PatchBranchHistories instead.
+  rpc PushCommits(PushCommitsRequest) returns (PushCommitsResponse);
```
saquibmian, line 52: I struggled to name these RPCs appropriately. Open to suggestions.

I also put this here instead of `BranchService` as I feel it's `Branch`-centric, but open to other suggestions.

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +118,117 @@ message ListBranchesResponse {
...
+  // A request to push a Commit to a Branch.
+  message Value {
+    // The Branch to push the Commit to.
+    BranchRef branch_ref = 1 [(buf.validate.field).required = true];
+    // The ID of the Commit to push to the Branch.
+    //
+    // This Commit will become the latest Commit on the Branch.
+    //
+    // If the Commit already exists on the Branch, an error is returned.
+    string commit_id = 2 [
```
saquibmian, line 132: `Commit`s don't have a `name`, I think that's why there's no `CommitRef`. That leaves two ways to reference a `Commit`: `id` or `digest`. We use `id` everywhere else (`Tag`, `Branch`, `VCSCommit`), so I stuck with that.

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -193,41 +193,28 @@ message CreateCommitsRequest {
...
-  // currently exist, a new Tag will be created for each name.
-  repeated string tag_names = 4 [(buf.validate.field).repeated.items = {
-    string: {max_len: 250}
-  }];
-  // Associated VCS commit information.
-  //
-  // If there are already VCSCommits on the associated Module with a given hash, this
-  // will result in an error. Otherwise, a new VCSCommit is created.
-  repeated AssociatedVCSCommit associated_vcs_commits = 5;
+  message Value {
```
saquibmian, line 196: This was primarily hoisted into a `Value` so that we can batch create `Commits` for multiple `ModuleSet`s in one RPC. Without this, I am forced to process one `ModuleSet` at a time.

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -193,41 +193,28 @@ message CreateCommitsRequest {
...
+    // A commit will be created for each ModuleNode.
+    repeated ModuleNode module_nodes = 1 [(buf.validate.field).repeated.min_items = 1];
+    // Blobs for the FileNodes referenced by module_nodes that are not present on the server.
+    //
+    // Only Blobs that were returned as missing from GetMissingBlobDigests need to be sent.
+    // Other Blobs already exist on the server, and will be ignored.
+    //
+    // If a FileNode within module_nodes has a Blob that is not on the server, and is not
+    // within missing_blobs, an error will be returned.
+    repeated buf.registry.storage.v1beta1.Blob missing_blobs = 2;
```
saquibmian, line 209: I think this should be hoisted out of the `Value`s? It's common to all of them.

buf/registry/module/v1beta1/branch_service.proto
bufdev: The `BranchService` relates to the `Branch` entity, the `CommitService` relates to the `Commit` entity. You are operating on a `Commit`, therefore it should be in the `CommitService`.

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +117,117 @@ message ListBranchesResponse {
...
+  // The requests to push a Commit to a Branch.
+  //
+  // Commits will be pushed to Branches in the order provided.
+  repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
+}
+
+message PushCommitsResponse {
+  message Value {
+    // The Branch to which the Commit was pushed.
+    Branch branch = 1 [(buf.validate.field).required = true];
```
bufdev, line 145: Why is this needed?

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +117,117 @@ message ListBranchesResponse {
...
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.max_len = 250
+    ];
+    // The ID of another Commit on another Branch from which to import Branch history. The history
+    // of any Branch that contains this Commit will be imported into the newly created Branch, up to
+    // and including this Commit.
+    //
+    // This is primarily designed for use in forking a Branch from an existing Branch instead of
+    // providing the new Branch's entire history.
+    optional string import_from_commit_id = 3 [(buf.validate.field).string.uuid = true];
```
bufdev, line 171: Why does this have field presence?

saquibmian: Would we prefer to just treat empty value as not set instead? I.e., only use field presence when the default value isn't sufficient?

buf/registry/module/v1beta1/branch_service.proto
saquibmian: My mental model is that I am operating on a `Branch` by putting a `Commit` on its history. The `Commit` is unaffected by this, nothing about it changes, but the `Branch` changes (it has a new latest commit).

Does that make sense, or am I thinking about this wrong?

akshayjshah: FWIW, my mental model of this is the same as Saquib's. This is an operation on a branch, not a commit.

buf/registry/module/v1beta1/commit_service.proto
bufdev: I want to understand the call pattern of this a bit better - how will I end up with `list<pair<list<module>, list<vcs_commit>>>`? Ie what is the repeated pair between `list<module> and list<vcs_commit>` for? Are we uploading multiple workspaces at once/would someone concievably want to do this?

buf/registry/module/v1beta1/commit_service.proto
bufdev: I'd agree - no matter what we do with `Value` (see above), this should be top-level in the request.

buf/registry/module/v1beta1/commit_service.proto
saquibmian: > Are we uploading multiple workspaces at once

Potentially, yes. 

Concretely, when we look at a Git repository, we have a Git commit with a workspace(ModuleSet) on it. Let's say I have several Git commits to upload workspaces for. In the current API, I can upload the entire workspace, with associated Git commits (i.e., squashing by identical digests), in one shot. But I can't upload multiple workspaces across Git commits where they are not identical by digest (i.e., I can't squash them) in a single shot. Batching it in the way proposed allows us to receive multiple workspaces that are not identical. 

This is not a requirement, I can always just fanout requests instead. But I figure we can't change this API later, so if we don't batch today, we are locked into that (unless we add another RPC later).

bufdev:
I updated with some edits - for ease of reference, here's the entire diff: https://github.com/bufbuild/registry-proto/compare/6d72e67e43c93f5011ca3943b073719141b61528...ad64ae80cef6ede1f5fd01cf30e66717fec8f978

Some main points:

- RPCs are now `Push` and `Rebase`.
- The concept of "head commit" now replaces the concept of "latest commit" - latest implies time, while "head" has clearer meaning.
- `PushRequest.Value` now takes mulltiple `commit_ids`, thereby batching a single `Branch` update into a single `Value`. This gets rid of potential duplicates, and makes it possible to have the response length match the request length. I think it's also just clearer for the caller as to what is happening.
- `Push` and `Rebase` now just return `Branches`, which are the only entity that actually gets modified I believe.
- `Operation`, `Operation.CreateBranch`, `Operation.UpdateBranch` are now the names of the operations on `RebaseRequest`. `CreateBranch` and `UpdateBranch` now nested inside `Operation`.
- `missing_blobs` moved top-level.
- Various other naming updates.
- Protobuf validation put in place for repeated fields.
- `BUF_VERSION` updated to v1.28.1 to make sure we are incorporating `PROTOVALIDATE` rules (I can't remember what the commit points to, and good to update anyways).
- Some extra documentation spec'ing out behavior.

Please double-check that I didn't miss anything and that the comments make sense and reflect the correct behavior, but assuming I didn't miss anything this is good to go from my end (but please triple-check).

bufdev (approved):

buf/registry/module/v1beta1/commit_service.proto
bufdev: OK makes sense.

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +115,132 @@ message ListBranchesResponse {
...
+message PushRequest {
+  message Value {
+    // The Branch to push the Commits to.
+    //
+    // If this is a name reference, and the named Branch does not exist, it will
+    // be created and the given Commits will be the first Commits on the Branch.
+    BranchRef branch_ref = 1 [(buf.validate.field).required = true];
+    // The ID of the Commits to push to the Branch, in the order they should be added.
+    //
+    // This Commit will become the head Commit on the Branch.
```
nicksnyder, line 128: Docs need updating now that this is a repeated field
```suggestion
    // The last Commit will become the head Commit on the Branch.
```

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +115,132 @@ message ListBranchesResponse {
...
+    //
+    // If the Commit already exists on the Branch, an error is returned.
```
nicksnyder, line 130: Docs need updating now that this is a repeated field
```suggestion
    // If any of these Commits already exist on the Branch, an error is returned.
```

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +115,132 @@ message ListBranchesResponse {
...
+        string: {uuid: true}
+      }];
+    }
+
+    oneof value {
+      option (buf.validate.oneof).required = true;
+      // Create a new branch.
+      CreateBranch create_branch = 1;
+      // Update an existing branch.
+      UpdateBranch patch_branch = 2;
```
nicksnyder, line 225: @bufdev should this variable be called `update_branch` to be consistent with the type?

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +115,132 @@ message ListBranchesResponse {
...
+      ];
+      // The ID of the Commit for which to start history from. The history of the branch will
+      // reflect the history as if you called ListCommitHistory for this Commit ID.
+      //
+      // If this is empty, the created Branch will only have the Commits specified in
+      // add_commit_ids.
+      //
+      // This is primarily designed for use in forking a Branch instead of
+      // providing the new Branch's entire history.
+      string start_commit_id = 3 [(buf.validate.field).string.uuid = true];
```
nicksnyder, line 173: @bufdev would this be better named `base_commit_id` to mirror the idea of a base commit in Git (and also better parallel the concept of `head` that we already use)?

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +115,132 @@ message ListBranchesResponse {
...
+      }
+    ];
+  }
+  // The Commits to push with the Branches they should be pushed to.
+  //
+  // All Values should have unique BranchRefs, that is no two Values should have
+  // a BranchRef that refers to the same Branch. An error will be returned if any
+  // two BranchRefs refer to the same Branch.
+  repeated Value values = 1 [(buf.validate.field).repeated.min_items = 1];
+}
```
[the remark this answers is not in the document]
saquibmian: We do not, but not supporting batching is inconsistent with the rest of the schema. Every RPC today supports batching, except the `List*` RPCs (which in theory could support batching trivially).

buf/registry/module/v1beta1/branch_service.proto
bufdev: Yes, I knew I'd miss something

buf/registry/module/v1beta1/branch_service.proto
bufdev: I could buy that - @akshayjshah thoughts?

buf/registry/module/v1beta1/branch_service.proto
bufdev: Yes

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -79,9 +78,11 @@ message ResolveCommitsRequest {
   //   - If a Commit is referenced, this just references that specific commit.
   //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
   //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
-  //   - Is a Branch is referenced, this is interpreted to mean the latest Commit on the Branch.
+  //   - Is a Branch is referenced, this is interpreted to mean the head Commit on the Branch.
   //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
   //     Digests referencing unreleased Commits cannot be referenced.
+  //
+  // TODO: what is "latest released Commit" if we now have "head Commit"?
```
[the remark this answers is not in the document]
bufdev: Yes we need to re-evaluate this

buf/registry/module/v1beta1/branch_service.proto
```diff
@@ -98,3 +115,132 @@ message ListBranchesResponse {
...
+        (buf.validate.field).string.uuid = true
+      ];
+      // The IDs of the Commits to remove from the Branch.
+      //
+      // Commits are removed before any Commits are added from add_commit_ids.
+      //
+      // Note that it is valid to have one of the remove_commit_ids be equal to the
+      // expected_head_commit_id. If this is the case, the Branch will have a different
+      // head Commit that add_commit_ids will be applied to.
+      repeated string remove_commit_ids = 3 [(buf.validate.field).repeated.items = {
```
unmultimedio, line 205: > Note that it is valid to have one of the remove_commit_ids be equal to the expected_head_commit_id. If this is the case, the Branch will have a different head Commit that add_commit_ids will be applied to.

This comment implies that we support removing commits _at the middle_ of the history, which is not true with the spec we're proposing. When removing, it's always commits at the tail of the branch, so it's not "_valid_ that one of the `remove_commit_ids` is the `expected_head_commit_id`", if we're removing commits, "the head is _required_ to be present as the last item in this array".

nicksnyder: You are right the current doc is a bit underspecified. In addition to documenting the point you made, I think we should also require/document that the order of commits matters (it is the order that they will be popped off the branch and at each step if the head commit id doesn't match the id trying to be popped off, the op fails).

saquibmian: @unmultimedio is correct, the ID of the head Commit _must_ be present in `remove_commit_ids`.

To @nicksnyder's point, this doc used to say

> These commits must appear in order at the end of the Branch's history.

I think we should keep defining the expected order of `commits_ids_to_remove`. I would also expect this to match the order in which they appear in Branch history, not the order in which they will be removed. The order in which they will be removed is not defined by the user, it's defined by the system. We simply care that the state we're asserting is valid, the actual order of how history is updated is based on some business rules that may change over time.

In other words, we receive a request that says:

> my history looks like `(..., A, B, C]`, I would like it to be `(..., A, D]`

We only need to validate that request matches our current state, and follow our business rules to carry out that request. We can _define_ how this will happen, but I wouldn't want the user to drive this (and for us to validate the user's expectations of how this will happen).

buf/registry/module/v1beta1/branch.proto
```diff
@@ -60,13 +60,13 @@ message Branch {
   // Only one Branch per module will be marked as the release Branch.
   // TODO: enum?
   bool is_release = 9 [(buf.validate.field).required = true];
-  // The id of the latest Commit created on the Branch.
-  string latest_commit_id = 10 [
+  // The id of the head Commit created on the Branch.
+  string head_commit_id = 10 [
     (buf.validate.field).required = true,
```
saquibmian, line 65: I noticed this is required. Does this imply that an empty Branch does not exist, or should we relax this constraint?

buf/registry/module/v1beta1/commit_service.proto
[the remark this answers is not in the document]
nicksnyder: I think we should try to land this PR and save this bigger TODO for a followup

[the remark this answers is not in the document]
saquibmian: Perfect, I'll leave this here for now and we can clean up/formalize "release" after this.

nicksnyder:
Please clean up PR description to reflect the actual implementation/names we landed on (and remove the TODO list). Basically make the PR description what the merged commit description should look like.

buf/registry/module/v1beta1/branch.proto
nicksnyder: Unless this causes a problem elsewhere, it makes sense to me that an empty branch doesn't exist because this is true in Git (a branch pointer has to point to a commit).

saquibmian: I updated the docs to reflect this.

nicksnyder:
Akshay lgtm'd in Slack and Peter told me live today that minor changes are fine and don't need his re-review, so I am going to click the merge button on this.

The only changes that happened since Peter took a look were docs changes and idempotency (which Akshay reviewed).

PR #44, opened 2023-11-30
Add filtering options for ListRequestUsers

emcfarlane:
Adds two new filtering options for `UserService/ListRequestUsers`:
- `has_types`: only return users of types.
- `has_states`: only return users of states.

Required behaviour for API consumers.


buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -115,12 +115,12 @@ message ListCommitHistoryRequest {
   //   - If a Digest is referenced, history is started at the latest released Commit that has this Digest.
   //     Digests referencing unreleased Commits cannot be referenced.
   ResourceRef resource_ref = 3 [(buf.validate.field).required = true];
+  // Whether to sort the Commits in descending order.
```
emcfarlane, line 118: Fixed typo on `sort` and reordered numbers for consistency.

bufdev (approved):

PR #45, opened 2023-12-04
Update `Commit.id` to not validate as a UUID

saquibmian:
`Commit.id` <=> `RepositoryCommit.name`, which is a dashless UUID.

This also implies that `ResourceRef.value.id` is not a UUID.

TODO:
- [ ] This is now inconsistent across the API surface. Should we drop the UUID validation on all IDs? Ideally our IDs are opaque to users and we are not expressing their shape in the API. Alternatively, we add `Commit.name` back and `Commit.id` <=> `RepositoryCommit.id` and `Commit.name` <=> `RepositoryCommit.name`.

### DIFF
--- a/buf/registry/module/v1beta1/resource.proto
+++ b/buf/registry/module/v1beta1/resource.proto
@@ -95,7 +95,7 @@ message ResourceRef {
   oneof value {
     option (buf.validate.oneof).required = true;
     // The id of the resource.
-    string id = 1 [(buf.validate.field).string.uuid = true];
+    string id = 1;
     // The fully-qualified name of the resource.
     Name name = 2;
   }
--- a/buf/registry/module/v1beta1/resource.proto
+++ b/buf/registry/module/v1beta1/resource.proto
@@ -95,7 +95,7 @@ message ResourceRef {
   oneof value {
     option (buf.validate.oneof).required = true;
     // The id of the resource.
-    string id = 1;
+    string id = 1 [(buf.validate.field).string.pattern = "^([0-9a-fA-F]{32})|([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})$"];
     // The fully-qualified name of the resource.
     Name name = 2;
   }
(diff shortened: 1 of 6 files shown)


nicksnyder:
The inconsistency between some IDs are UUIDs and other aren’t is a bit unfortunate.

I think we *should* leverage Protovalidate for basic sanity checking of input as much as possible, so I don't think we should remove `(buf.validate.field).string.uuid = true` from id fields where it is valid.

For fields that can't use the build-in `(buf.validate.field).string.uuid = true` rule, we shouldn't just remove it, we should replace it with actual validation. It sounds like they are in fact UUIDs without the dashes, so let's validate that.

saquibmian:
> It sounds like they are in fact UUIDs without the dashes, so let's validate that.

Are we sure we want to go down that path? We are exposing the shape of our IDs, and we cannot change this later.

In the meantime, I'll update the validation rules to validate a dashless UUID.

buf/registry/module/v1beta1/resource.proto
```diff
@@ -95,7 +95,7 @@ message ResourceRef {
   oneof value {
     option (buf.validate.oneof).required = true;
     // The id of the resource.
-    string id = 1 [(buf.validate.field).string.uuid = true];
```
nicksnyder, line 98: I assume we could write a CEL expression that essentially validates this field is either a UUID or whatever the validation is for a commit id (sounds like it is a UUID without dashes).

nicksnyder:
> Are we sure we want to go down that path? We are exposing the shape of our IDs, and we cannot change this later.

This is an interesting philosophical question. My two arguments in favor are:
1. In practice, it is nice to be able to rely on protovalidate to validate basic things like this.
2. In theory, even if we remove these validations from the schema, Hyrum's law will apply (people will see that in practice we are using UUIDs everywhere).

Therefore might as well put it in the schema and accept that if we are compelled to change this later it would be a breaking change in that sense.

saquibmian:
Closing this PR, the BSR will take care to translate `name` to `id` and back as needed. We discussed this as a team and would like to prioritize the best API possible. Decision doc [here](https://www.notion.so/bufbuild/Decision-what-do-with-Commit-id-7926f1bc61774429bd4db6abf357b3f9?pvs=4).

Issue #18, opened 2023-11-06
Move spdx_license_id out of Commit and create LicenseService or similar


bufdev:

nicksnyder:
spdx_license_id was removed from Commit here: https://github.com/bufbuild/registry-proto/pull/50/files#r1439632570.

Need new way to retrieve it.

PR #50, opened 2023-12-22
Refactor

bufdev:


buf/registry/module/v1beta1/digest.proto
```diff
@@ -0,0 +1,48 @@
...
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/module/v1beta1";
+
+// A digest of a Commit's content.
+//
+// A digest represents all content for a single Commit, including its .proto files, documentation
+// files, license files, and dependencies.
+//
+// TODO: should this be named something other than Digest, to make sure it is clear that this
+// is not related to the digest of a single File? This was originally named "CommitDigest" but
+// if we're trying to move away from Commits, this seems wrong.
```
saquibmian, line 30: If we're moving away from Commits, what entity refers to a collection of files (proto, license, and documentation files) and dependencies to other collections of files? The Commit is the only thing that encapsulates this today. In the CLI we call this the Module, but Module refers to something else in the API.

buf/registry/module/v1beta1/commit.proto
```diff
@@ -48,17 +53,31 @@ message Commit {
     (buf.validate.field).required = true,
     (buf.validate.field).string.uuid = true
   ];
-  // The digest of the Commit's contents.
+  // The Digest of the Commit's contents.
```
saquibmian, line 56: The term `content` here was confusing to me because in this context it includes metadata about dependencies to other commits but in another context (DownloadService) it refers to files only.

saquibmian: Nevermind, I misread that Content includes deps. Ignore this.

saquibmian (commented):
WIP review. I started looking at UploadService but had some questions that I should probably understand before I go further. In UploadService it's strange that we upload Content but end up advancing Labels as well.

The main question I have: To what extent are we trying to move off of Commits? Currently they are the only entity that expresses dependencies between sets of files. I see we're introducing a new "Content" concept, which is interesting, but we should perhaps model that explicitly? Something like this?

```proto
message Content {
  Digest digest = 1; // Contents are addressable by Digest, but we can give them IDs for consistency
  string module_id = 2;
  repeated File files = 3;
  repeated Digest dependencies = 4;
}
```

And then is a Commit just a Content on a particular Label's history? I.e., has other metadata like create_time, author, but is otherwise not used unless you need that metadata? If we're going down that path, we'd have a `ContentService` and callers can upload Content, but Labels would be advanced separately. I.e., creating a Content is not analogous to creating a Commit. This could be undesirable, though, orphaned Content, etc.

If the intention is to upload Content and create Commits in one go, then I think we should operate at the Commit level, not at the Content level. The caller would create a _Commit_, not a Content. But that's Commit-centric, taking me back to my original question...

buf/registry/module/v1beta1/download_service.proto
```diff
@@ -0,0 +1,167 @@
...
+    //
+    // See the documentation on ResourceRef for resource resolution details.
+    //
+    // Once the resource is resolved, the following content is returned:
+    //   - If a Module is referenced, the content of the default Label is returned.
+    //   - If a Label is referenced, the content associated with this Label is returned.
+    //   - If a Commit is referenced, the content for the Commit is returned.
+    //   - If a Digest is referenced, the content the latest Commit with this Digest is returned.
+    // TODO: Is "latest Commit with this Digest" correct? If you're requesting a Digest, it likely
+    // doesn't matter.
```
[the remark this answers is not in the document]
nicksnyder: To the extent that a digest can refer to multiple commits, and if we are returning commit specific info in the response, then we do need to define which commit we will select in this case (and latest makes sense to me).

buf/registry/module/v1beta1/download_service.proto
```diff
@@ -0,0 +1,167 @@
...
+    //
+    // If include_dep_content is true, all deps will also have associated Content.
+    repeated Commit deps = 2;
+  }
+  // The references in the same order as requested.
+  repeated Reference references = 1 [(buf.validate.field).repeated.min_items = 1];
+  // The Contents of the references, ordered by commit_id.
+  //
+  // If include_dep_content is true, all deps will also have associated Content. Otherwise,
+  // only the Commits associated with the Referencves will have associated Content.
```
nicksnyder, line 162:
```suggestion
  // only the Commits associated with the References will have associated Content.
```

buf/registry/module/v1beta1/resource.proto
```diff
@@ -43,24 +42,20 @@ message ResourceRef {
...
+  //       is of the form "typeString:hexValue", where typeString is the lowercase of the Type
+  //       value with TYPE_ stripped, and hexValue is the hex-encoded value of the digest.
+  //     - If there is a conflict between names across resources (for example, there is a Commit id
+  //       and Label name of the same value), the following order of precedence is applied:
   //       - Commit
-  //       - VCSCommit
-  //       - Tag
-  //       - Branch
+  //       - Label
+  //       - Digest
```
nicksnyder, line 58: I would think Digest needs to have precedence over a label. The whole point of a digest value is it always points to the same content, but if I can create a label that redefines a digest value, that is kind of like a hash collision attack.

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -0,0 +1,91 @@
...
+    repeated ModuleRef dep_module_refs = 2;
+    // The labels to associate with the Commits created by the upload.
+    //
+    // If an id is set, this id must represent a Label that already exists and is
+    // owned by the Module. The Label will point to the newly-created Commits for the References,
+    // or will be updated to point to the pre-existing Commit for the Reference.
+    //
+    // Note that at least one label must be referenced. If you want to reference the default
+    // label, do a request against the ModuleService to determine the ID of the default label,
+    // and reference it here.
```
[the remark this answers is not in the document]
bufdev: I'm not sure what you mean here, you can upload content and advance the label no problem, you can't atomically change what the default label is (ie from `master` to `main`), upload content, and change its `commit_id`, but I don't think that's desirable. Regardless, this was removed for other reasons.

buf/registry/module/v1beta1/download_service.proto
```diff
@@ -0,0 +1,167 @@
...
+    // the response). If a directory "foo/bar" is specified but this directory has no files,
+    // this is considered to be a non-match.
+    //
+    // This field also interacts with file_types - if file_types is set, a path only matches
+    // if it is also of the file type, and if there are no matching paths for the given FileTypes,
+    // an error is returned unless paths_not_allow_exist is set.
+    //
+    // Note that paths may reference dependency paths if include_dep_files is set.
+    //
+    // The path must be relative, and connect contain any "." or ".." components
```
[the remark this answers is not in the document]
bufdev: Done

buf/registry/module/v1beta1/label.proto
```diff
@@ -17,94 +17,105 @@ syntax = "proto3";
...
-    // The fully-qualified name of the Tag.
+    // The fully-qualified name of the Label.
     Name name = 2;
   }
 }
+
+// A reference to a Label scoped to a Module, either an id or a name.
+//
+// This is used in requests.
+message ScopedLabelRef {
```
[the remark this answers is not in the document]
bufdev: Sounds right

buf/registry/module/v1beta1/download_service.proto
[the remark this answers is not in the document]
bufdev: Sounds good

buf/registry/module/v1beta1/download_service.proto
```diff
@@ -0,0 +1,167 @@
...
+service DownloadService {
+  // Download the contents of multiple Modules, Labels, Digests, or Commits.
+  //
+  // Content consists of:
+  //   - The Files: .proto files, license files, and documentation files.
+  //   - The dependencies.
+  //
+  // You may specify whether to additionally download the Content of dependencies via the
+  // include_dep_content field.
+  rpc Download(DownloadRequest) returns (DownloadResponse) {
```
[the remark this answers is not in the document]
bufdev: Ive pushed updates with what I think is the final shape

buf/registry/module/v1beta1/resource.proto
bufdev: Done

buf/registry/module/v1beta1/graph.proto
```diff
@@ -0,0 +1,40 @@
...
+package buf.registry.module.v1beta1;
+
+import "buf/registry/module/v1beta1/commit.proto";
+import "buf/registry/priv/extension/v1beta1/extension.proto";
+import "buf/validate/validate.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/module/v1beta1";
+
+// A dependency graph.
+message Graph {
```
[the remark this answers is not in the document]
bufdev: We can talk about it, but leaving as-is for now - maybe it makes more sense here, but could be a slippery slope to naming `Label` as `CommitLabel`, etc.

buf/registry/module/v1beta1/download_service.proto
```diff
@@ -0,0 +1,169 @@
...
+import "buf/registry/module/v1beta1/commit.proto";
+import "buf/registry/module/v1beta1/digest.proto";
+import "buf/registry/module/v1beta1/file.proto";
+import "buf/registry/module/v1beta1/resource.proto";
+import "buf/validate/validate.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/module/v1beta1";
+
+// Download contents.
+service DownloadService {
```
emcfarlane, line 28: What about `DownloadService` and `UploadService` combined into a single `ContentService`?

buf/registry/module/v1beta1/module.proto
```diff
@@ -65,10 +65,14 @@ message Module {
     (buf.validate.field).string.max_len = 255,
     (buf.validate.field).ignore_empty = true
   ];
-  // The id of the Branch which releases are generated from.
-  string release_branch_id = 10 [
+  // The name of the default label of the Module.
+  //
+  // This label may not yet exist. When a Module is created, it has not Commits, and Labels
```
nicksnyder, line 70:
```suggestion
  // This label may not yet exist. When a Module is created, it has no Commits, and Labels
```

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -17,341 +17,135 @@ syntax = "proto3";
...
-  //   - If a Tag is referenced, history is started at the Commit associated with the Tag.
-  //   - If a VCSCommit is referenced, history is started at the Commit associated with the VCSCommit.
-  //   - Is a Branch is referenced, history is started at the head Commit on the Branch.
-  //   - If a Digest is referenced, history is started at the latest released Commit that has this Digest.
-  //     Digests referencing unreleased Commits cannot be referenced.
+  // See the documentation on Ref for resource resolution details.
   //
-  // TODO: what is "latest released Commit" if we now have "head Commit"?
+  // Once the resource is resolved, the following Commits are listed:
+  //   - If a Module is referenced, all Commits for the Module are returned.
```
nicksnyder, line 84: This should probably be scoped to the default label instead of mixing commits across all labels
```suggestion
  //   - If a Module is referenced, all Commits for the default Label of the Module are returned.
```

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -17,341 +17,135 @@ syntax = "proto3";
...
+  //
+  // See the documentation on ResourceRef for resource resolution details.
+  //
+  // Resoluation is as follows:
+  //   - If a Module is referenced, the Commit of the default Label is returned.
+  //   - If a Label is referenced, the Commit of this Label is returned.
+  //   - If a Commit is referenced, this Commit is returned.
+  //   - If a Digest is referenced, the latest Commit with this Digest is returned.
+  // TODO: Is "latest Commit with this Digest" correct? If you're requesting a Digest, it likely
+  // doesn't matter.
```
nicksnyder, line 141: Looks like duplicated doc
```suggestion
```

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -17,341 +17,135 @@ syntax = "proto3";
...
-message CreateCommitsResponse {
-  // The created Commits, in the order of the nodes given via module_nodes.
+  // Once the resource is resolved, this reference is interpeted to refer to the following Commit:
+  //   - If a Module is referenced, this is interpreted to mean the latest released Commit on the Module.
+  //   - If a Commit is referenced, this just references that specific commit.
+  //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
+  //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
+  //   - Is a Branch is referenced, this is interpreted to mean the head Commit on the Branch.
+  //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
+  //     Digests referencing unreleased Commits cannot be referenced.
```
nicksnyder, line 128: The current proposal is not to have "released" be a property of a commit. Here is what I think the docs should be:

```suggestion
  // Once the resource is resolved, this reference is interpreted to refer to the following Commit:
  //   - If a Module is referenced, this is interpreted to mean the latest Commit on the default Label of the Module.
  //     If the default Label for this module has governance flow enabled, the latest Commit that has been approved will
  //     be returned.
  //   - If a Commit is referenced, this just references that specific commit.
  //   - If a Tag is referenced, this is interpreted to mean the Commit associated with the Tag.
  //   - If a VCSCommit is referenced, this is interpreted to mean the Commit associated with the VCSCommit.
  //   - Is a Branch is referenced, this is interpreted to mean the head Commit on the Branch.
  //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
```

nicksnyder (commented):
This batch of comments is about resolving commits.

One thing I am worried/confused about is that we currently have 4 different RPCs for fetching commits, and it is unclear if all of these are necessary.
* ListLabelHistoryRequest
* GetCommits
* ListCommits
* ResolveCommits

buf/registry/module/v1beta1/vcs_commit.proto
```diff
@@ -16,124 +16,49 @@ syntax = "proto3";
...
+    (buf.validate.field).required = true,
+    (buf.validate.field).enum.defined_only = true
+  ];
+  // The VCS hash of the Commit.
+  string hash = 2 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[0-9a-fA-F]{40}$"
+  ];
+  // The URL of of the commit within the VCS.
+  string url = 3 [(buf.validate.field).string.uri = true];
```
nicksnyder, line 63: @akshayjshah and I debated whether this should be a per-commit URL or a per-module URL. The current proposal in Notion is for this to be a per-module URL (so the current implementation deviates from the proposal, and if we keep this we should update the proposal). The distinction here is relevant if/when enterprises move code hosts (which does happen and is not uncommon). If we store this on a per-commit basis, and if a customer updates their code host but doesn't keep their old code host active and doesn't setup redirects, then we will end up with a bunch of dead links. Having a module level URL template would make this easy to remedy. 

Having said all that.
* Maybe it is ok to have old commits point to dead links?
* Maybe it is ok to see if enterprises keep old code hosts running or setup redirects in practice?
* Maybe it is ok we punt on this until there is an actual enterprise with an actual problem, and we could always add a module level override in the future (less config now)?

Semantically, all we really care to capture is a link back to the code host. Maybe this is better named CodeHostLink instead of VCS commit. Another argument for this is that URL isn't really a property of a VCS commit.

If we are going down the "embed url per commit" route, I think we should consider not having this structure at all and instead just having a single `string code_host_url` property on commit and call it a day.

buf/registry/module/v1beta1/download_service.proto
bufdev: I split them up because even though we can use authz for this, exposing read-only as a service (ie just download) is something I think we'll want to do externally on such a common basis, that its worth keeping it as a standalone service.

buf/registry/module/v1beta1/commit_service.proto
bufdev: This was mean to be deleted - deleting

buf/registry/module/v1beta1/vcs_commit.proto
bufdev: I am fine with whatever, was just coding up the proposal :-) we aren't implementing this part now anyways, we can adjust this pre-v1 to whatever you'd like, I can see both arguments (but would probably lean into per-module, now thinking about it)

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -17,341 +17,45 @@ syntax = "proto3";
...
-  //   - Is a Branch is referenced, this is interpreted to mean the head Commit on the Branch.
-  //   - If a Digest is referenced, this is interpreted to mean the latest released Commit that has this Digest.
-  //     Digests referencing unreleased Commits cannot be referenced.
+  // Resolution is as follows:
+  //   - If a Module is referenced, the Commit of the default Label is returned.
+  //   - If a Label is referenced, the Commit of this Label is returned.
+  //   - If a Commit is referenced, this Commit is returned.
   //
-  // TODO: what is "latest released Commit" if we now have "head Commit"?
+  // TODO: It's a little incongruent to have a Get request take a ResourceRef.
```
saquibmian, line 44: I actually think the primary use case of a `ResourceRef` _is_ a Get request. `ResourceRef` is really only needed because of the "opaque string" concept. In mutational cases, we should request the most specific thing possible (Commits ID if deps, Label IDs/names if labels to advance, etc.).

saquibmian (commented):
Threw a few more comments on, overall download/upload look cleaner!

buf/registry/module/v1beta1/commit.proto
```diff
@@ -48,17 +54,17 @@ message Commit {
...
+  Digest digest = 5 [(buf.validate.field).required = true];
   // The id of the User that created this Commit on the BSR.
   string created_by_user_id = 6 [
     (buf.validate.field).required = true,
     (buf.validate.field).string.uuid = true
   ];
-  // The SPDX license ID of the associated license, if any.
-  //
-  // To retrieve the specific license file, use the ContentService.
-  string spdx_license_id = 7 [(buf.validate.field).string.max_len = 255];
```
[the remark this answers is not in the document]
bufdev: https://github.com/bufbuild/registry-proto/issues/18

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -0,0 +1,93 @@
...
+    // the module component of the ResourceRef should be specified, and specifying any other
+    // part of the ResourceRef will result in an error.
+    //
+    // See the documentation on ResourceRef for resource resolution details.
+    //
+    // Resolution is as follows:
+    //   - If a Module is referenced, the Commit of the default Label is used.
+    //   - If a Label is referenced, the Commit of this Label is used.
+    //   - If a Commit is referenced, this Commit is used.
+    repeated ResourceRef dep_resource_refs = 2;
```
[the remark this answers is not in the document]
bufdev: Modules are referenced without commits when they are `References`:

```
In this case, only the module component of the ResourceRef should be specified.
```

This can either be a ModuleRef or a CommitRef, effectively. We figured we might as well allow LabelRefs for non-buf-CLI customers as well. The upload comment doesn't require resolution - it may have different callers than the buf CLI.

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -0,0 +1,93 @@
...
+  }
+  // A set of content for a given reference.
+  message Content {
+    // The Module of the content.
+    ModuleRef module_ref = 1 [(buf.validate.field).required = true];
+    // The Files of the content.
+    //
+    // This will consist of the .proto files, license files, and documentation files.
+    repeated File files = 2 [(buf.validate.field).repeated.min_items = 1];
+  }
```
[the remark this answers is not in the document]
bufdev: Added a TODO

buf/registry/module/v1beta1/vcs_commit.proto
```diff
@@ -16,124 +16,49 @@ syntax = "proto3";
...
+// service and moves the functionality to a different module. We want the BSR commits created
+// from A and C to link back to the correct git commits and PR discussions.)
+//
+// Multiple BSR commits may have the same VCS hash, but different content. This should only
+// happen when users incorrectly hand-specify the VCS hash — and since commits are immutable and
+// can’t be deleted, we don’t want a manually-created commit to forever block a later,
+// fully-correct commit.
+message VCSCommit {
+  // The type of VCS that the Commit originated from.
+  VCSType type = 1 [
```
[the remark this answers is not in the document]
bufdev: OK - that's an implementation detail, we likely should be storing it, however.

buf/registry/module/v1beta1/vcs_commit.proto
```diff
@@ -16,124 +16,49 @@ syntax = "proto3";
...
+    (buf.validate.field).required = true,
+    (buf.validate.field).enum.defined_only = true
+  ];
+  // The VCS hash of the Commit.
+  string hash = 2 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[0-9a-fA-F]{40}$"
```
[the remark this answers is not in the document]
bufdev: When we support things other than git, we can drop it.

bufdev:
Merging as-is - improvements can happen as we go, this isn't final.

PR #53, opened 2024-01-03
Change delete to archive for Labels

bufdev:


buf/registry/module/v1beta1/module.proto
```diff
@@ -65,11 +65,13 @@ message Module {
...
     (buf.validate.field).ignore_empty = true
   ];
-  // The name of the default label of the Module.
+  // The name of the default Label of the Module.
   //
   // This label may not yet exist. When a Module is created, it has no Commits, and Labels
   // must have a Commit, so this Label is not created when a Module is created. Additionally,
   // a User may modify the name of the default Label without this Label yet being created.
+  //
+  // This may not point to an archived Label.
```
nicksnyder, line 74: I actually don't think this is true. This can point to any arbitrary string, and the next push will either create the label if it doesn't exist or unarchive it if it is unarchived. Ditto for other instances

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -40,22 +40,25 @@ service LabelService {
...
   rpc CreateOrUpdateLabels(CreateOrUpdateLabelsRequest) returns (CreateOrUpdateLabelsResponse) {
     option idempotency_level = IDEMPOTENT;
   }
-  // Delete existing Labels.
+  // Archive existing Labels.
   //
-  // This operation is atomic. Either all Labels are deleted or an error is returned.
-  rpc DeleteLabels(DeleteLabelsRequest) returns (DeleteLabelsResponse) {
+  // This operation is atomic. Either all Labels are archived or an error is returned.
+  rpc ArchiveLabels(ArchiveLabelsRequest) returns (ArchiveLabelsResponse) {
```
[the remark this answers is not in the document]
bufdev: No you can't, there's no `archived` field on `CreateOrUpdateLabels`

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -92,6 +95,8 @@ message ListLabelsRequest {
   //   - If a Module is referenced, all Labels for the Module are returned.
   //   - If a Label is referenced, this Label is returned.
   //   - If a Commit is referenced, all Labels for the Commit are returned.
+  //
+  // If an archived Label is referenced but include_archived is false, this will return an error.
```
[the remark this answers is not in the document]
bufdev: Just remove this comment

buf/registry/module/v1beta1/module.proto
bufdev: OK sounds good, can update comment

buf/registry/module/v1beta1/resource.proto
```diff
@@ -32,6 +32,9 @@ option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/r
 //
 // ResourceRefs can only be used in requests, and only for read-only RPCs, that is
 // you should not use an arbitrary reference when modifying a specific resource.
+//
+// ResourceRefs cannot reference archived Labels. The only way to retrieve an archived Label
+// is to use LabelService.GetLabels.
```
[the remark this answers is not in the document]
bufdev: No it can't, there is no such filter

buf/registry/module/v1beta1/label_service.proto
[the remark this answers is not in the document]
doriable: I see, right, you can only change what a label is pointing to, and then there is the behaviour of auto-unarchiving.

buf/registry/module/v1beta1/label_service.proto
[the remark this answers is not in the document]
doriable: But this field would be needed to support getting a set of labels that includes both archived and unarchived labels for a given resource ref (e.g. commit) right?

doriable (commented):
Can push updates since you're on the road, @bufdev, but I think our comments are generally resolved.

buf/registry/module/v1beta1/module.proto
```diff
@@ -65,11 +65,15 @@ message Module {
...
-  // The name of the default label of the Module.
+  // The name of the default Label of the Module.
   //
   // This label may not yet exist. When a Module is created, it has no Commits, and Labels
   // must have a Commit, so this Label is not created when a Module is created. Additionally,
   // a User may modify the name of the default Label without this Label yet being created.
+  //
+  // This could also be the name of an archived label. An important note, an archived label
+  // will not be considered a valid ResourceRef for resolving a Module. Archived labels
+  // can only be retrieved using LabelService.GetLabels.
```
doriable, line 76: Updated this comment, however there is still pending conversation on archived labels for `ListLabels`, will need to change this part of the comment based on that.

buf/registry/module/v1beta1/module.proto
doriable: I talked to @nicksnyder and updated the comment.

buf/registry/module/v1beta1/label_service.proto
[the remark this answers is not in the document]
nicksnyder: I think I see the dilemma here. We kind of want to enforce by default that archived labels can't be used to resolve anything important, so this change is proposing that `ResourceRef` can't reference an archived label.

The consequence of this is that ListLabels cannot return archived labels because the request parameter here is a ResourceRef.

The future use-case @akshayjshah is interested in is being able to display all the labels that ever pointed to that commit, even archived ones. In this future world, we would want either:
1. This method to be able to support returning not only archived labels, but also historical labels (not just the labels currently pointing to a commit). The three things we would need:
  1. Add `include_archived` to include archived labels in response (this can be added later)
  2. Add `include_historical` (or equivalent) to include labels that used to point to this commit (this can be added later)
  3. Either allow ResourceRef to reference archived labels, or use a different type here, potentially a `oneof` ResourceRef and LabelRef. This would be a breaking change from current behavior, so if we want this we need to do it now.
2. Create a new RPC.

If we are ok with creating a new RPC in the future for this use case if we need it, then I think we can leave the API as-is and remove this comment.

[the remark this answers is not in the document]
doriable: I think for option 1, if we are using `include_*` fields, which would have backwards-compatible defaults, we can also implement later down the road. I think I'm happy removing this comment and we can revisit this when we need a solution.

doriable (approved):
Approving this since I believe all comment threads are wrapped up. I think deferring the `ListLabels` filtering is very reasonable (or in the future create a new RPC).

buf/registry/module/v1beta1/label_service.proto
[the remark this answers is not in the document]
nicksnyder: Chatted with Doria live. I was mixed up about needing (1)(iii). We don't need to change the request to accomplish this use case because we will be requesting a commit id (which is valid) and returning archive labels (which is valid).

This means (1)(i) and (1)(ii) are sufficient, and we can do both of those in the future in a backward compatible way. Resolving this thread.

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -132,6 +135,8 @@ message ListLabelHistoryRequest {
   // If empty, the first page is returned.
   string page_token = 2 [(buf.validate.field).string.max_len = 4096];
   // The Label to list history for.
+  //
+  // This can reference archived Labels.
```
nicksnyder, line 139: I just noticed that we allow querying the history of an archived label. I think this is right/necessary, but also seems to contradict a comment above about `If a Label is archived, the API considers it to not exist, except when calling LabelService.GetLabels`.

I am wondering if we need to update that comment to say GetLabels is is the only endpoint that can _return_ an archived label.

nicksnyder: just to be consistent
```suggestion
  // This may reference archived Labels.
```

doriable: I think that we make it clear that archived labels can only be retrieved by `LabelService.GetLabels` in our documentation. When an RPC takes 1+ label(s) as an input, it always accepts archived labels (e.g. `CreateOrUpdateLabel`, `ListLabelHistory`, etc.) so I think this is helpful clarifying documentation, but does not create any inconsistencies to the mental model.

buf/registry/module/v1beta1/label.proto
```diff
@@ -37,32 +37,37 @@ message Label {
   google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
   // The last time the Label was updated on the BSR.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
+  // Whether or not the Label was archived.
+  //
+  // If a Label is archived, the API considers it to not exist and will not return archived
+  // labels, except when calling LabelService.GetLabels.
```
nicksnyder, line 43:
```suggestion
  // Labels, except when calling LabelService.GetLabels.
```

buf/registry/module/v1beta1/module.proto
```diff
@@ -65,11 +65,15 @@ message Module {
...
+  // The name of the default Label of the Module.
   //
-  // This label may not yet exist. When a Module is created, it has no Commits, and Labels
+  // This Label may not yet exist. When a Module is created, it has no Commits, and Labels
   // must have a Commit, so this Label is not created when a Module is created. Additionally,
   // a User may modify the name of the default Label without this Label yet being created.
+  //
+  // This could also be the name of an archived label. An important note, an archived label
+  // will not be considered a valid ResourceRef for resolving a Module. Archived labels
+  // can only be retrieved using LabelService.GetLabels.
```
nicksnyder, line 76:
```suggestion
  // This could also be the name of an archived Label. An important note, an archived Label
  // will not be considered a valid ResourceRef for resolving a Module. Archived Labels
  // can only be retrieved using LabelService.GetLabels.
```

nicksnyder (approved):

buf/registry/module/v1beta1/module_service.proto
```diff
@@ -135,6 +135,8 @@ message CreateModulesRequest {
     // The name of the default Label of the Module.
     //
     // If not set, the default Label will be named "main" upon creation.
+    //
+    // This may not point to an archived Label.
```
nicksnyder, line 139: @doriable didn't we decide that default_label_name CAN point to an archived label?

https://github.com/bufbuild/registry-proto/pull/53/commits/07bf7b6f8820f872019e1d675c4deda101ab6e26

buf/registry/module/v1beta1/module_service.proto
```diff
@@ -166,9 +168,11 @@ message UpdateModulesRequest {
       (buf.validate.field).string.max_len = 255,
       (buf.validate.field).ignore_empty = true
     ];
-    // The name of the default label of the Module.
+    // The name of the default Label of the Module.
+    //
+    // This Label may not yet exist.
     //
-    // This label may not yet exist.
+    // This may not point to an archived Label.
```
nicksnyder, line 175: ditto

buf/registry/module/v1beta1/module_service.proto
doriable: Oh, not sure how this got revert :S let me add it back.

PR #55, opened 2024-01-09
Remove VCSCommit and add Commit.source_control_url

nicksnyder:
The BSR doesn't need to be in the business of storing source control information--we don't semantically need a vcs commit hash for example and we don't care what type of source control is being used.

As a convenience for users, we DO want to store a link back to the relevant commit in source control so that a user can find relevant info (e.g. commit description, PR discussion, authors, approvers, etc.).

This change simplifies the data model to only store what we need, which is a single source control URL per commit.


bufdev (approved):
I worry a bit that we're going to have a feature in the future where we do want the hash, and it doesn't cost us much to store, and it's not too expensive for us to provide it, BUT with that said I'm OK with this

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -81,8 +80,11 @@ message UploadRequest {
     //
     // If no labels are referenced, the default Label for the Module is used.
     repeated ScopedLabelRef scoped_label_refs = 6;
-    // VCS Commit information.
-    VCSCommit vcs_commit = 7;
+    // The URL for the source control commit that is associated with the BSR Commit's Contents.
```
saquibmian, line 83:
```suggestion
    // The URL for the source control commit to associate with the Commit created by the upload.
```

buf/registry/module/v1beta1/commit.proto
```diff
@@ -65,6 +64,9 @@ message Commit {
     (buf.validate.field).required = true,
     (buf.validate.field).string.uuid = true
   ];
-  // VCS commit information.
-  VCSCommit vcs_commit = 7;
+  // The URL for the source control commit that is associated with the BSR Commit's Contents.
```
saquibmian, line 67:
```suggestion
  // The URL for the source control commit that is associated with the Commit.
```

buf/registry/module/v1beta1/upload_service.proto
saquibmian: Related, there's only one Commit created per Module, right? The documentation for `scoped_label_refs` uses `Commits`, which is misleading (one might read that as a Commit per ScopedLabelRef). Might be worth cleaning that up to ensure we're clear and consistent here.

buf/registry/module/v1beta1/commit.proto
```diff
@@ -65,6 +64,9 @@ message Commit {
...
+  //
+  // BSR users can navigate to this link to find source control information that is relevant to this BSR Commit
+  // (e.g. commit description, PR discussion, authors, approvers, etc.).
+  string source_control_url = 7 [(buf.validate.field).string.uri = true];
```
pkwarren, line 71: Do we want to ensure the scheme here is `https` (and `http` if we want to be lenient), or do we need to handle custom schemes depending on the SCM type?

akshayjshah:
> I worry a bit that we're going to have a feature in the future where we do want the hash, and it doesn't cost us much to store, and it's not too expensive for us to provide it, BUT with that said I'm OK with this

I felt the same way, but I'm having a really hard time figuring out what that situation might be.

I don't think we can have the BSR call Github or git directly, because it'll open up a lot of questions about accessing files beyond the schemas. If the BSR can't access git or Github, I can't imagine anything git-specific that we'd be able to do. Multiple BSR commits can have the same VCS hash (in theory), so we decided not to allow resolving BSR commits by VCS hash.

Given all that, the only use case I can really imagine is search. I think we can enable that pretty easily by parsing the URL, breaking the path into segments, and throwing away segments less than N characters/containing non-hex characters.

On the plus side, this approach clearly gets the BSR out of the source control business: we just link to source control via a URI, and we only use it for providing context to users.

buf/registry/module/v1beta1/commit.proto
nicksnyder: accepted

buf/registry/module/v1beta1/upload_service.proto
nicksnyder: I cleaned this up

buf/registry/module/v1beta1/commit.proto
nicksnyder: Eh, I don't see a need to enforce a particular URL scheme here. The BSR itself doesn't really care what this url is at all really. We are just going to render a link to it that a user can click on. If some company has some use case for deep linking into some native app that is on every developer's machine, I see no reason to stand in their way.

PR #56, opened 2024-01-16
Set create_by_user_id on Commit to optional

emcfarlane:
On user deletion `create_by_user_id` is no longer available. Remove the required option from the validation.


bufdev:
Can you add a comment saying why it's optional, ie the above?

PR #64, opened 2024-02-05
CreateOrUpdateLabels* becomes CreateLabels*

nicksnyder:
Our dependency resolution algorithm prefers to select commits by the latest creation date, but that yields non-sensical results if we allow labels to get updated to point to older commits (which the current API permits).

For example, it is currently possible to use CreateOrUpdateLabels to construct a label history like [C1, C2, C1]. Semantically C1 should be the "latest" commit, but since it has a timestamp older than C2, our dependency resolution algorithm would prefer C2.

To avoid this scenario, we should not expose an endpoint to explicitly update the commit a label is pointing to. Instead, the only way to update a label is to use UploadService. This allows users to push the content they want to push (which can be identical content to a previous commit) and we can mint a new commit for them with a current created_at timestamp. In other words, the commit history from the example above would be [C1, C2, C3] where the content of C3 is identical to the content of C1.

The alternative to this is to keep the current API but document it as being an error to attempt to update a label to point to a commit with a created_at timestamp that is earlier than the existing commit the label points to. This alternative is implemented here: https://github.com/bufbuild/registry-proto/pull/66

### DIFF
--- a/buf/registry/module/v1beta1/label_service.proto
+++ b/buf/registry/module/v1beta1/label_service.proto
@@ -38,13 +38,11 @@ service LabelService {
   rpc ListLabelHistory(ListLabelHistoryRequest) returns (ListLabelHistoryResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
-  // Create or update Labels on a Module.
+  // Create Labels on a Module.
   //
-  // If the Label does not exist, it will be created.
-  // If the Label was archived, it will be unarchived.
-  //
-  // This operation is atomic. Either all Labels are created/updated or an error is returned.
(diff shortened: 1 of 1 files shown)


unmultimedio:
I lean more towards #66 to keep supporting [the `buf tag` command](https://buf.build/docs/reference/cli/buf/beta/registry/tag/create#usage) (which probably will evolve to `buf label` or similar, that adds labels to existing commits.

nicksnyder:
Fair enough. If everyone is fine with https://github.com/bufbuild/registry-proto/pull/66 then we can close this PR

bufdev:
I think we should do #66.

PR #66, opened 2024-02-05
Update CreateOrUpdateLabel RPC docs

unmultimedio:
Add the constraint of newer commit `create_time` timestamps when updated a Label pointer.

### DIFF
--- a/buf/registry/module/v1beta1/label_service.proto
+++ b/buf/registry/module/v1beta1/label_service.proto
@@ -171,7 +171,8 @@ message CreateOrUpdateLabelsRequest {
     // The id of the Commit to associate with the Label.
     //
     // If the Label already existed, the Label will now point to this Commit, as long as this Commit
-    // is newer than the Commit that the Label is currently pointing to.
+    // is newer than the Commit that the Label is currently pointing to, otherwise an error is
+    // returned.
     // If the Label was archived, it will be unarchived.
     string commit_id = 2 [
       (buf.validate.field).required = true,
--- a/buf/registry/module/v1beta1/label_service.proto
+++ b/buf/registry/module/v1beta1/label_service.proto
@@ -42,6 +42,8 @@ service LabelService {
   //
   // If the Label does not exist, it will be created.
   // If the Label was archived, it will be unarchived.
+  // If the Label already existed, the Commit in the request has to be newer than the Commit that
+  // the Label is currently pointing to, otherwise an error is returned.
   //
   // This operation is atomic. Either all Labels are created/updated or an error is returned.
   rpc CreateOrUpdateLabels(CreateOrUpdateLabelsRequest) returns (CreateOrUpdateLabelsResponse) {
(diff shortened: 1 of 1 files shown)


nicksnyder:
We need to either merge this or https://github.com/bufbuild/registry-proto/pull/64

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -170,7 +170,8 @@ message CreateOrUpdateLabelsRequest {
     LabelRef label_ref = 1 [(buf.validate.field).required = true];
     // The id of the Commit to associate with the Label.
     //
-    // If the Label already existed, the Label will now point to this Commit.
+    // If the Label already existed, the Label will now point to this Commit, as long as this Commit
+    // is newer than the Commit that the Label is currently pointing to.
```
nicksnyder, line 174: I think we should be explicit that it is an error to attempt to update a label to point to an older commit.

```suggestion
    // If the Label already existed, the Label will now point to this Commit as long as this Commit
    // is newer than the Commit that the Label is currently pointing to, otherwise an error is returned.
```

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -170,7 +170,9 @@ message CreateOrUpdateLabelsRequest {
     LabelRef label_ref = 1 [(buf.validate.field).required = true];
     // The id of the Commit to associate with the Label.
     //
-    // If the Label already existed, the Label will now point to this Commit.
+    // If the Label already existed, the Label will now point to this Commit, as long as this Commit
```
bufdev, line 173: This documentation should be copied and/or moved to the RPC documentation  - we should explain that you cannot update labels to earlier commits

unmultimedio: Copied the docs to the RPC.

bufdev (approved):

nicksnyder (approved):

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -42,6 +42,8 @@ service LabelService {
   //
   // If the Label does not exist, it will be created.
   // If the Label was archived, it will be unarchived.
```
nicksnyder, line 44: wait, it will only be unarchived if a new commit is pushed, right?

unmultimedio: Good question. If label is `[A, B, C]` and archived, and I send a request to update it to `D`, I'd expect it to become `[A, B, C, D]` _and_ unarchived.

nicksnyder: agree, but if you try to update the label to point to a commit that is older than C, it should fail (which means that it should not become unarchived). The current docs here are written in a way that make it seem like it will always become unarchived.

saquibmian: Later on down it says
> This operation is atomic. Either all Labels are created/updated or an error is returned.

So I read this as "if the request fails, no state is changed".

saquibmian (approved):

buf/registry/module/v1beta1/label_service.proto
nicksnyder: ok, as long as we are all on the same page, that is what I care about. good enough to merge

PR #67, opened 2024-02-12
Make registry field more lenient

jchadwick-buf:
Protovalidate doesn't yet have a rule that covers what we need for the registry field. We internally rely on the field being allowed to contain an address with a port, not just a hostname; while protovalidate does have an address validator that allows either an IP or a hostname, it lacks an address with optional port validator. Rather than try to write a complicated regular expression, I think it's best to just allow the field to contain any string, which will eventually be validated implicitly by trying to use it as an address anyways.


bufdev (approved):

rodaine:
Adding a constraint for this: https://github.com/bufbuild/protovalidate/pull/164

You'll still need to use a custom constraint if you want the port to be optional (defaulting to the port being required for the standard constraint).

PR #69, opened 2024-02-13
Support policy checks

bufdev:


buf/registry/module/v1beta1/label_service.proto
```diff
@@ -157,12 +157,56 @@ message ListLabelHistoryRequest {
...
+  message Value {
+    // The Commit.
+    Commit commit = 1 [(buf.validate.field).required = true];
+    // The Commit's breaking state.
+    //
+    // This field will only be returned if breaking change detection is enabled.
+    //
+    // Breaking change detection is an enteprise-only feature - contact us to
+    // learn more!
+    BreakingState breaking_state = 2 [(buf.validate.field).enum.defined_only = true];
```
akshayjshah, line 195: Should this be repeated? If we keep it singular, then client code needs to encode the hierarchy of breaking categories (e.g., that breaking at the `PACKAGE` level implies breaking at the `FILE` level). If we make it repeated, the server can respond with _all_ the categories (e.g., `[PACKAGE, FILE]`) and the clients can stay dumb.

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -157,12 +157,56 @@ message ListLabelHistoryRequest {
...
+    // The Commit's review state.
+    //
+    // The field will only be returned if review flow is enabled.
+    //
+    // Review flow is an enteprise-only feature - contact us to
+    // learn more!
+    ReviewState review_state = 3 [(buf.validate.field).enum.defined_only = true];
+  }
```
akshayjshah, line 203: Will we want a timestamp for the last change to review state?

buf/registry/module/v1beta1/label_service.proto
bufdev: Hmm I don't know, I don't feel too strongly either way. Arguments against `repeated`:

- Hierarchy is pretty-well understood and agreed upon since the beginning of  `buf`.
- Making this `repeated` enables states that are incorrect (even if we never respond to them) - what does it mean to have `[NONE, WIRE]`? What does it mean to have `[WIRE, FILE]` but not `WIRE_JSON` or `PACKAGE`?

I might lean slightly against `repeated`, but don't know.

buf/registry/module/v1beta1/label_service.proto
bufdev: I don't know, we might, should we add now or defer until we need it?

buf/registry/module/v1beta1/label.proto
```diff
@@ -71,6 +72,30 @@ message Label {
     (buf.validate.field).required = true,
     (buf.validate.field).string.uuid = true
   ];
+
+  // Whether or not governance is enabled for this Label.
+  //
+  // Governance is an enterprise-only feature - contact us to learn more!
+  bool governance_enabled = 10;
```
akshayjshah, line 79:
```suggestion
  bool review_enabled = 10;
```

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -96,7 +97,9 @@ message ListLabelsRequest {
   // Once the resource is resolved, the following Labels are listed:
   //   - If a Module is referenced, all Labels for the Module are returned.
   //   - If a Label is referenced, this Label is returned.
```
akshayjshah, line 99: I think we probably need to handle filtering out the label the user supplied (if it doesn't match the review states filter below).

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -106,6 +109,10 @@ message ListLabelsRequest {
   // TODO: We are purposefully not making the default the zero enum value, however
   // we may want to consider this.
   Order order = 4 [(buf.validate.field).enum.defined_only = true];
+  // Only return Labels that have these governance review states.
+  //
+  // By default, Labels with any governance review state are returned.
+  repeated LabelReviewState governance_review_states = 5;
```
akshayjshah:
```suggestion
  repeated LabelReviewState review_states = 5;
```

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -157,12 +164,26 @@ message ListLabelHistoryRequest {
 }
 
 message ListLabelHistoryResponse {
+  message Value {
+    // The Commit.
+    Commit commit = 1 [(buf.validate.field).required = true];
+
+    // The Label's review state.
+    //
+    // This field will only be set if governance is enabled for the Label.
```
akshayjshah, line 173: Think the docs here should clarify that this is the state of the latest commit for that label.

buf/registry/module/v1beta1/label.proto
```diff
@@ -71,6 +72,30 @@ message Label {
...
   ];
+
+  // Whether or not governance review is enabled for this Label.
+  //
+  // Governance is an enterprise-only feature - contact us to learn more!
+  bool review_enabled = 10;
+  // The current review state of the Label.
+  //
+  // This field will only be set if governance is enabled.
+  LabelReviewState review_state = 11 [(buf.validate.field).enum.defined_only = true];
```
nicksnyder, line 83: I don't understand what this field is for. Is this the review state of the commit pointed to by `commit_id`? If so, that can never be `_REJECTED`.

Is this the review state of the most recent commit on the label (Which may be any state, but may be the state of a commit that is different than the commit pointed to by `commit_id`)?

nicksnyder: In other words, review state is only a property of a Label plus a Commit, but since this field is on a Label, it is unclear which Commit the review state applies to

nicksnyder: I see this was answered below in the docs on LabelReviewState. My bad.

buf/registry/module/v1beta1/label_service.proto
saquibmian: Is this not the review state of this particular Commit in this history of this Label? If so, it is odd to use `LabelReviewState` here, the docs on that reference "last commit". I would expect a nested enum like `CommitReviewState`.

nicksnyder:
I am taking at pass at this and will push up a proposed change.

buf/registry/module/v1beta1/label_service.proto
nicksnyder: Any particular reason? I don't think we should special case this.

buf/registry/module/v1beta1/label_service.proto
nicksnyder: Seems good to have since it tells us when a commit was "approved" on a label (which may be later than when it was pushed). Looks like it got added to the diff already, so going to resolve this thread.

nicksnyder (commented):
I renamed LabelReviewState to CommitReviewState because review states are more closely bound to Commits than Labels (they are 1:1 with the former and Many:1 with the latter). Also cleaned up some docs. 

Before merging, I think we should decide how we want to consistently refer to this feature because that might impact names like `review_enabled`, but I started a Slack thread about that.

buf/registry/module/v1beta1/label_service.proto
nicksnyder: Yeah, I agree with Saquib here.

I pushed a commit to rename `LabelReviewState` to `CommitReviewState`.

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -96,7 +96,9 @@ message ListLabelsRequest {
   // Once the resource is resolved, the following Labels are listed:
   //   - If a Module is referenced, all Labels for the Module are returned.
   //   - If a Label is referenced, this Label is returned.
-  //   - If a Commit is referenced, all Labels for the Commit are returned.
+  //   - If a Commit is referenced, all Labels for the Commit are returned. Note that this will
+  //     not return any Labels where the Commit is currently rejected or pending, it will only
+  //     return Labels for Commits that are passed or accepted.
```
saquibmian, line 101: What about blocked commits? I imagine we want to treat them similar to pending?

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -157,12 +163,22 @@ message ListLabelHistoryRequest {
 }
 
 message ListLabelHistoryResponse {
+  message Value {
+    // The Commit.
+    Commit commit = 1 [(buf.validate.field).required = true];
+
+    // The GovernanceWorkflowState for this Commit on this Label.
+    //
+    // This field will only be set if governance workflow was enabled on this Label when this Commit was pushed.
```
saquibmian, line 172: What about blocked commits in a Label that doesn't have governance workflow enabled?

buf/registry/module/v1beta1/label.proto
```diff
@@ -71,6 +72,47 @@ message Label {
...
     (buf.validate.field).string.uuid = true
   ];
+  // Whether or not governance workflow is enabled for this Label.
+  //
+  // Governance workflow is an enterprise-only feature - contact us to learn more!
+  bool governance_workflow_enabled = 10;
+  // The governance workflow state of the Commit the Label points to.
+  //
+  // This field will only be set if governance workflow was enabled on this Label when this Commit was pushed.
+  GovernanceWorkflowState governance_workflow_state = 11;
```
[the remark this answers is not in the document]
nicksnyder: I am not familiar with our conventions around unspecified enum values, but my assumption is that we treat that like an invalid state and would never return it in practice. Otherwise, yes, a label can only point to a commit that either doesn't have a GovernanceWorkflowState or one that does with a status of either accepted or passed.

buf/registry/module/v1beta1/label_service.proto
nicksnyder: yes, I pushed a f3d9cb2f024fdc12041b7e0550b7a8eba919b2c0 to update docs

buf/registry/module/v1beta1/label_service.proto
nicksnyder: updated docs to clarify, ptal

nicksnyder (approved):
Ok, after iterating on this a bit and discussing some things around naming with Mary, the current state of this PR is my proposal.

"Governance workflow" can/will be the umbrella of features that include policy checks, but also include things that aren't necessarily related to approving changes (e.g. audit logs, permissions, etc). An analogy here is that this is "Compliance".

"Policy checks" are specifically checks that run to determine if a commit can be placed on a label's history. An analogy here is "CI jobs" (which must pass before merging).

PolicyChecksState is the overall state of all policy checks for a given Commit x Label. Currently we have one implied check: breaking change detection. In the future we could have more (e.g. lint), but there would still be a single top level status (the one already proposed here). In the future we could add lower level status for individual policy checks (e.g. so you can know the specific status for breaking change detection, or lint, or otherwise), but we can add that to PolicyChecksState if/when we need it.

Other changes
* Expanded the types to support dependencies between policy checks (which we need because of workspace push).
* Standardize on "Approved" instead of "Accepted" since the former seem more likely what we would want to use in the UI (we already use this today, and this is what code review tools use so it seems like the right term of art)

I am out next week, so if everyone else is ok with this, please go ahead and merge.

buf/registry/module/v1beta1/label.proto
```diff
@@ -71,6 +72,55 @@ message Label {
...
+  // This field will be set if either:
+  //   - Policy checks were enabled on this Label when this Commit was pushed.
+  //   - This Commit transitively depends on a Commit that has a PolicyChecksState.
+  PolicyChecksState policy_checks_state = 11;
+}
+
+// The state of a Commit's policy checks for a particular Label.
+//
+// Policy checks are an enterprise-only feature - contact us to learn more!
+message PolicyChecksState {
```
bufdev, line 90: I *really dislike* the plural naming of this - even just `PolicyCheckState` would be better, but I don't know if that fits into your model.

buf/registry/module/v1beta1/label.proto
```diff
@@ -71,6 +72,55 @@ message Label {
...
+  // The status of the policy checks.
+  PolicyChecksStatus status = 1 [(buf.validate.field).enum.defined_only = true];
+
+  // The time the status was last updated.
+  google.protobuf.Timestamp status_updated_at = 3 [(buf.validate.field).required = true];
+
+  // The ids of Commits that have policy checks blocking these policy checks.
+  //
+  // This field will only be set if the status is blocked.
+  repeated string blocked_on_commit_ids = 2 [(buf.validate.field).string.uuid = true];
```
bufdev, line 100: This repeated field is in `Commit`, and could be unbounded - if I have 10000 of these, are all of them going to appear on a `Commit`? I'm generally hesitant to have `repeated` fields on Resources, wondering if this is required.

buf/registry/module/v1beta1/label.proto
nicksnyder: I don't feel strongly about this and an open to naming this `PolicyCheckState`.

`PolicyCheckState` implies to me a single overall policy check based on potentially multiple other checks (currently just breaking change detection, but in the future, there could be more checks that roll up into this overall state).

`PolicyChecksState` implies that we may be running multiple independent policy checks (currently only breaking).

Both names work for me so I am fine just renaming this.

buf/registry/module/v1beta1/label.proto
nicksnyder: We are actually still debating whether we need this or not, so it might go away entirely. In the event that we keep it...

> This repeated field is in Commit
This isn't meant to be a field on `Commit` it is info that needs to be bound to a union of a Commit and a Label (which IIRC we have a table for on the backend but this is not explicitly modeled in the API)

> could be unbounded - if I have 10000 of these, are all of them going to appear on a Commit?

I think the natural bound for this field would be the number of modules that are in a single workspace (because that is the only circumstance where this blocking state would be relevant). Does that constraint this to a scale you are comfortable with?

This is mostly for display purposes in the UI, so I think designing a separate service/entity for paging would be overkill. If this ends up being a problem, I would rather go in the other direction of just documenting that it will contain a maximum of 10-100 items (or whatever limit you are comfortable with). The UI purpose is to show you want might need to be reviewed to get this commit unblocked so it is sufficient to only show some of them if there are a large number.

buf/registry/module/v1beta1/label.proto
maryelizbeth: I'm fine w/ `PolicyCheckState` as long as we are fine with rolling future checks into this (e.g. linting); so basically +1 to Nick's opinion

nicksnyder: Renamed in f42ba0c440a700b0730f091137aa372e210b60e0

nicksnyder (approved):

buf/registry/module/v1beta1/label.proto
nicksnyder: removed this from the PR for now. Will add back in a future PR if we decide to implement blocked status

PR #74, opened 2024-02-15
Make `dep_refs.commit_id` optional

unmultimedio:
This is only populated if the dependency commit already exists, but for codependent modules being pushed, `commit_id` should not be set.

### DIFF
--- a/buf/registry/module/v1beta1/upload_service.proto
+++ b/buf/registry/module/v1beta1/upload_service.proto
@@ -42,7 +42,7 @@ message UploadRequest {
     //
     // If the ModuleRef refers to a Module that has associated Content, this field should *not*
     // be set, and setting it is an error.
-    string commit_id = 2 [(buf.validate.field).string.uuid = true];
+    optional string commit_id = 2 [(buf.validate.field).string.uuid = true];
   }
   // Content to upload for a given reference.
   message Content {


buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -42,7 +42,7 @@ message UploadRequest {
     //
     // If the ModuleRef refers to a Module that has associated Content, this field should *not*
     // be set, and setting it is an error.
-    string commit_id = 2 [(buf.validate.field).string.uuid = true];
+    optional string commit_id = 2 [(buf.validate.field).string.uuid = true];
```
saquibmian, line 45: `optional` means something very specific, namely that this field will have field presence. I don't think we need presence here, the default value (empty string) should suffice, and it does in various other places in the API: https://github.com/bufbuild/registry-proto/blob/8054b81b8e70e42eeed9ffd50fd1ad29af78ab1a/buf/registry/legacy/federation/v1beta1/resource.proto#L77-L78

We could update the documentation to clarify this.

unmultimedio: Sure. @emcfarlane agrees with you. I don't have a preference tbh, I thought it doesn't make sense to send and empty value here and instead the requester shouldn't set it, but alright I'll do the change and update docs. 👍🏼

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -40,9 +40,12 @@ message UploadRequest {
...
     // The commit_id of the Commit, if this is referencing a Commit that already exists.
     //
-    // If the ModuleRef refers to a Module that has associated Content, this field should *not*
-    // be set, and setting it is an error.
-    string commit_id = 2 [(buf.validate.field).string.uuid = true];
+    // If the ModuleRef refers to a Module that has associated Content, this field should be empty,
+    // and setting a value is an error.
+    string commit_id = 2 [
+      (buf.validate.field).string.uuid = true,
+      (buf.validate.field).ignore_empty = true
```
saquibmian, line 47: Ah I see, I didn't realize empty string (i.e., unset) did not pass validation. https://github.com/bufbuild/protovalidate/blob/8074ae2f42417345ef103d83fb62e4245010715d/proto/protovalidate/buf/validate/validate.proto#L2898-L2902
I'm not sure what's the way we want to go here, so I'll defer to Peter.

unmultimedio: For reference:
- some URLs use this model: https://github.com/bufbuild/registry-proto/blob/97279adc5b464d8296b2731ad22ddb2f2cfc253c/buf/registry/legacy/federation/v1beta1/upload_service.proto#L93-L101
- some URLs have both presence and `ignore_empty`: https://github.com/bufbuild/registry-proto/blob/97279adc5b464d8296b2731ad22ddb2f2cfc253c/buf/registry/module/v1beta1/module_service.proto#L165-L170
- some descriptions use just optional: https://github.com/bufbuild/registry-proto/blob/97279adc5b464d8296b2731ad22ddb2f2cfc253c/buf/registry/owner/v1beta1/organization_service.proto#L149-L150

It'd be nice to get them all to a consistent state. I think presence should be the way to go, unless the zero value is actually valid **and different** than unset, but for URLs and description that doesn't seem like the case. Wdyt @bufdev

bufdev (approved):

PR #75, opened 2024-02-20
Change usage of ignore_empty to IGNORE_IF_UNPOPULATED

bufdev:


rodaine:
In all these cases, IGNORE_DEFAULT is equivalent to IGNORE_EMPTY. Where you might want IGNORE_DEFAULT is an `optional string foo` in (say) an update request that allows zero-ing out a field on a resource (that's labeled with IGNORE_EMPTY).

bufdev:
Yea I know they're equivalent, I was just trying to standardize on a world where we wanted to have a recommended value of `Ignore` to use, and trying to model that behavior. `IGNORE_DEFAULT` seemed more like that?

Probably want to give people a recommendation in the `Ignore` docs too.

rodaine:
I don't think a recommendation makes sense, given that it's dependent on which semantics you want. The reason it has options is precisely because the meaning changes (in a subtle nuanced way, but still significant).

bufdev:
Changed to `IGNORE_IF_UNPOPULATED`

rodaine (approved):

PR #76, opened 2024-02-23
Move DepRefs and DigestType to top-level

bufdev:
This PR's primary purpose is to remove the need to specify dependencies between Contents on Upload, and to move external dependencies to the top-level of the UploadRequest. We do not need to manually specify dependencies between Contents, as this is implicitly done via .proto imports. We also want to make sure that for a given set of Contents, there is exactly one dependency Commit ID for a given Module. This matches what we have in Workspaces. The API will no longer enable uploading multiple Workspaces in one RPC, but we see no use case for this.

This PR also moves DigestType to the top-level of DownloadRequest, to match the other RPCs.

### DIFF
--- a/buf/registry/legacy/federation/v1beta1/upload_service.proto
+++ b/buf/registry/legacy/federation/v1beta1/upload_service.proto
@@ -104,6 +104,9 @@ message UploadRequest {
   //
   // Dependencies between Contents are implicit and do not need to be specified. The BSR will detect
   // dependencies between Contents via .proto imports.
+  //
+  // Commits should be unique by Module, that is no two dep_refs should have the same Module but
+  // different Commit IDs.
   repeated DepRef dep_refs = 2;
 }
 
--- a/buf/registry/legacy/federation/v1beta1/upload_service.proto
+++ b/buf/registry/legacy/federation/v1beta1/upload_service.proto
@@ -44,7 +44,7 @@ message UploadRequest {
   // A dependency of one or more references specified by Contents.
   //
   // Dependencies between Contents are implicit and do not need to be specified. The BSR will detect
-  // dependencies between Contenta via .proto imports.
+  // dependencies between Contents via .proto imports.
   message DepRef {
     // The commit_id of the dependency.
     string commit_id = 1 [
--- a/buf/registry/legacy/federation/v1beta1/upload_service.proto
+++ b/buf/registry/legacy/federation/v1beta1/upload_service.proto
@@ -41,36 +41,32 @@ service UploadService {
 // See the package documentation for more details. You should likely use buf.registry.module.v1beta1
 // and not this package.
 message UploadRequest {
-  // A dependency of Content, either referencing another Content message, or referencing
-  // a Commit that already exists.
+  // A dependency of one or more references specified by Contents.
+  //
+  // Dependencies between Contents are implicit and do not need to be specified. The BSR will detect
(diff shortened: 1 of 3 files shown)


buf/registry/legacy/federation/v1beta1/upload_service.proto
```diff
@@ -41,44 +41,40 @@ service UploadService {
 // See the package documentation for more details. You should likely use buf.registry.module.v1beta1
 // and not this package.
 message UploadRequest {
-  // A dependency of Content, either referencing another Content message, or referencing
-  // a Commit that already exists.
+  // A dependency of one or more references specified by Contents.
+  //
+  // Dependencies between Contents are implicit and do not need to be specified. The BSR will detect
+  // dependencies between Contenta via .proto imports.
```
saquibmian, line 47:
```suggestion
  // dependencies between Contents via .proto imports.
```

buf/registry/legacy/federation/v1beta1/upload_service.proto
```diff
@@ -41,44 +41,40 @@ service UploadService {
...
   message DepRef {
-    // The Module of the dep.
-    buf.registry.module.v1beta1.ModuleRef module_ref = 1 [(buf.validate.field).required = true];
```
unmultimedio, line 48: I think having both `ModuleRef` + `commit_id` helps making more explicit that I depend on a commit of _that_ module.

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -33,45 +33,27 @@ service UploadService {
 }
 
 message UploadRequest {
-  // A dependency of Content, either referencing another Content message, or referencing
-  // a Commit that already exists.
-  message DepRef {
-    // The Module of the dep.
-    ModuleRef module_ref = 1 [(buf.validate.field).required = true];
```
unmultimedio, line 40: Same here

buf/registry/legacy/federation/v1beta1/upload_service.proto
```diff
@@ -89,19 +85,29 @@ message UploadRequest {
...
+  // The dependencies of the references specified by Contents.
+  //
+  // This will include all transitive dependencies.
+  //
+  // Dependencies between Contents are implicit and do not need to be specified. The BSR will detect
+  // dependencies between Contents via .proto imports.
+  //
+  // Commits should be unique by Module, that is no two dep_refs should have the same Module but
+  // different Commit IDs.
+  repeated DepRef dep_refs = 2;
```
saquibmian, line 110: Is it an error to include a DepRef that the BSR prunes away? I.e., if a Commit is referenced here that is never imported (even transitively) by any Content.

buf/registry/legacy/federation/v1beta1/upload_service.proto
bufdev: That's duplicate information - a commit_id represents a Commit, and a Commit has a specific module. The rest of our API returns Commits, which have a module ID, as an example. Specifying duplicate information on an API request is an anti-pattern

buf/registry/legacy/federation/v1beta1/upload_service.proto
bufdev: Not an error I would say

saquibmian (approved):

unmultimedio (approved):

PR #77, opened 2024-02-23
Make dep commit ids unique

unmultimedio:


saquibmian (approved):

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -86,7 +86,10 @@ message UploadRequest {
   //
   // Commits should be unique by Module, that is no two dep_commit_ids should have the same Module but
   // different Commit IDs.
-  repeated string dep_commit_ids = 2 [(buf.validate.field).repeated.items.string.uuid = true];
+  repeated string dep_commit_ids = 2 [
+    (buf.validate.field).repeated.unique = true,
```
bufdev, line 90: We didn't use the unique validation throughout the API because it could be expensive - there's lot of other places we'd want to use it if we were going to be consistent. It's not actually enforcing the uniqueness we really need anyways - we need commits to be unique by module. A duplicate commit isn't actually an issue.

unmultimedio: Yeah I know the uniqueness is across modules (not commits), and I thought of making it at the code level, since I'm already looping through them anyway, but more of a conceptual question, is passing the same commit multiple times an error? May be not but I'm not sure.

bufdev: I don't think so - I think it's fine, we don't enforce this in other places, I'm not worried about it

PR #81, opened 2024-03-01
Add start_commit_id to ListLabelHistoryRequest

nicksnyder:
To view a diff for a commit on a label we need to be able to get the previous commit on that label. Added `start_commit_id` so we can fetch the history starting at the desired commit and using a page size of 2 can also get the previous commit.

### DIFF
--- a/buf/registry/module/v1/label_service.proto
+++ b/buf/registry/module/v1/label_service.proto
@@ -143,6 +143,11 @@ message ListLabelHistoryRequest {
   uint32 page_size = 1 [(buf.validate.field).uint32.lte = 250];
   // The page to start from.
   //
+  // The value may be a page_token returned from ListLabelHistoryResponse,
+  // or it may be a Commit id.
+  //
+  // If a Commit id, the returned history will start at that Commit.
+  //
   // If empty, the first page is returned.
   string page_token = 2 [(buf.validate.field).string.max_len = 4096];
   // The Label to list history for.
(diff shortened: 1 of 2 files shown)


buf/registry/module/v1/label_service.proto
```diff
@@ -143,6 +143,11 @@ message ListLabelHistoryRequest {
   uint32 page_size = 1 [(buf.validate.field).uint32.lte = 250];
   // The page to start from.
   //
+  // The value may be a page_token returned from ListLabelHistoryResponse,
+  // or it may be a Commit id.
+  //
+  // If a Commit id, the returned history will start at that Commit.
```
unmultimedio, line 149: Nit: Add a comment saying that if the passed Commit id is not part of the label history, this returns an error.

unmultimedio (approved):
For context: we already have the property that a commit can be in the label history just once, so this is reasonable to me. Wdyt @saquibmian ?

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -144,6 +144,12 @@ message ListLabelHistoryRequest {
   uint32 page_size = 1 [(buf.validate.field).uint32.lte = 250];
   // The page to start from.
   //
+  // The value may be a page_token returned from ListLabelHistoryResponse,
+  // or it may be a Commit id.
+  //
+  // If a Commit id, the returned history will start at that Commit.
```
bufdev, line 150: Special-casing page_token for a single endpoint makes me extremely nervous on a bunch of levels, can discuss

nicksnyder: Changed to start_commit_id

bufdev (approved):

PR #84, opened 2024-03-06
Fix field numbers for v1 ListLabelHistoryRequest

nicksnyder:
Field number 5 was inadvertently skipped in https://github.com/bufbuild/registry-proto/pull/80


bufdev (approved):

bufdev:
Breaking changes OK as we haven't deployed yet, overriding.

PR #87, opened 2024-03-06
Add only_content_changes to ListLabelHistoryRequest

nicksnyder:
This will make it possible to fetch only the commits on a label where the module content actually changed.

### DIFF
--- a/buf/registry/module/v1/label_service.proto
+++ b/buf/registry/module/v1/label_service.proto
@@ -178,9 +178,6 @@ message ListLabelHistoryResponse {
     Commit commit = 1 [(buf.validate.field).required = true];
     // The CommitCheckState for this Commit on this Label.
     CommitCheckState commit_check_state = 2 [(buf.validate.field).required = true];
-    // Whether or not the content of this Commit is different
-    // compared to the previous Commit in the history of this Label.
-    bool content_change = 3;
   }
   // The next page token.
   //
--- a/buf/registry/module/v1/label_service.proto
+++ b/buf/registry/module/v1/label_service.proto
@@ -167,9 +167,9 @@ message ListLabelHistoryRequest {
     (buf.validate.field).string.uuid = true,
     (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
   ];
-  // Whether or not to only return Commits where the content changed
-  // compared to the previous Commit in the history of this Label.
-  bool only_content_changes = 7;
+  // Only list Commits where the Digest has changed from the previous Commit in the
+  // history of this Label.
+  bool only_commits_with_changed_digests = 7;
 }
 
 message ListLabelHistoryResponse {
--- a/buf/registry/module/v1/label_service.proto
+++ b/buf/registry/module/v1/label_service.proto
@@ -175,6 +175,9 @@ message ListLabelHistoryResponse {
     Commit commit = 1 [(buf.validate.field).required = true];
     // The CommitCheckState for this Commit on this Label.
     CommitCheckState commit_check_state = 2 [(buf.validate.field).required = true];
+    // Whether or not this Commit represents a change in the Protobuf schema
+    // compared to the previous Commit on this Label.
+    bool protobuf_change = 3;
   }
   // The next page token.
   //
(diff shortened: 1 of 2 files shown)


buf/registry/module/v1/label_service.proto
```diff
@@ -167,6 +167,9 @@ message ListLabelHistoryRequest {
     (buf.validate.field).string.uuid = true,
     (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
   ];
+  // Whether or not to only return Commits that represent a change in the Protobuf schema
```
saquibmian, line 170: I think you mean this:
```suggestion
  // Whether or not to only return Commits that represent a change in Content
```

If we _do_ mean schema, what is included in that? Is a change to a comment also considered a schema change? Change to the license file?

buf/registry/module/v1/label_service.proto
```diff
@@ -175,6 +178,9 @@ message ListLabelHistoryResponse {
     Commit commit = 1 [(buf.validate.field).required = true];
     // The CommitCheckState for this Commit on this Label.
     CommitCheckState commit_check_state = 2 [(buf.validate.field).required = true];
+    // Whether or not the content of this Commit is different
+    // compared to the previous Commit in the history of this Label.
+    bool content_change = 3;
```
bufdev, line 183: Shouldn't you be able to just compare the digest of the previous commit in your response? This is duplicate info otherwise.

buf/registry/module/v1/label_service.proto
```diff
@@ -167,6 +167,9 @@ message ListLabelHistoryRequest {
     (buf.validate.field).string.uuid = true,
     (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
   ];
+  // Whether or not to only return Commits where the content changed
+  // compared to the previous Commit in the history of this Label.
+  bool only_content_changes = 7;
```
bufdev, line 172: This feels like a UI-only need that can be deduced from the information present in the API already - unless there's a strong reason otherwise (maybe there is), we purposefully designed registry-protoc to get out of the practice of making lots of API compromises in the name of a single call pattern

buf/registry/module/v1/label_service.proto
nicksnyder: whoops, you are correct we don't need this. Removed.

buf/registry/module/v1/label_service.proto
nicksnyder: The current API makes it possible to get this information, but not in a performant way in the general case. For repositories with a lot of commits that don't actually change the modules in them (i.e. monorepos with a lot of non-protobuf commits), clients will have to page through an arbitrary number of requests of this endpoint to find/fill one page of actual results to display to a user if the user only cares to see commits that only actually changed a module. This is why a filter flag here seems necessary.

We could drop this from the BSR design, but it seems like an important and valuable use case in general (similar to `git log -- [filename]`). I also think it would be a valuable flag to have on the CLI too when fetching the history for a label (but we don't need to do this now). Given all that, I don't really see this as a compromise in the API, we are just making it possible to satisfy a valid/valuable use case.

bufdev: Ok - I think I would name this field "skip_no_diff_commits", or something to do with "skip" and "commits" (open to other suggestions)

bufdev (approved):

buf/registry/module/v1beta1/label_service.proto
```diff
@@ -175,6 +175,9 @@ message ListLabelHistoryRequest {
     (buf.validate.field).string.uuid = true,
     (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
   ];
+  // Only list Commits where the Digest has changed from the previous Commit in the
+  // history of this Label.
+  bool only_commits_with_changed_digests = 8;
```
saquibmian, line 180: Which Digest type? The one requested? 

B4 digests include config files, so I suspect you mean B5, but one may think this is the Digest type from the request.

nicksnyder: Not sure, let's discuss in sync today. The intent here is to only show commits where observable changes happened (aka there is a diff, aka something material to the module changed), and we are using the digest as a way to express this. Either way, I would want make sure it is not possible to set this to true and return a history where two sequential commits have the same requested digest because that would be confusing.

nicksnyder: Decided on B5 comparison. For v1 this is the only digest anyway.

PR #88, opened 2024-03-08
Add name_query filter to ListLabels

nicksnyder:
This enables users to find matching labels if they only type in a substring of the label name (this is for the label selector in a module).

### DIFF
--- a/buf/registry/module/v1/label_service.proto
+++ b/buf/registry/module/v1/label_service.proto
@@ -116,7 +116,7 @@ message ListLabelsRequest {
   // TODO: Add custom CEL validation to validate the status field is one of DISABLED, PASSED, APPROVED.
   repeated CommitCheckStatus commit_check_statuses = 5 [(buf.validate.field).repeated.items.enum.defined_only = true];
   // Only return Labels with a name that contains this string.
-  string name = 6;
+  string name = 6 [(buf.validate.field).string.max_len = 250];
 }
 
 message ListLabelsResponse {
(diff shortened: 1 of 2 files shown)


buf/registry/module/v1/label_service.proto
```diff
@@ -115,6 +115,8 @@ message ListLabelsRequest {
   //
   // TODO: Add custom CEL validation to validate the status field is one of DISABLED, PASSED, APPROVED.
   repeated CommitCheckStatus commit_check_statuses = 5 [(buf.validate.field).repeated.items.enum.defined_only = true];
+  // Only return Labels with a name that contains this string.
+  string name = 6 [(buf.validate.field).string.max_len = 250];
```
nicksnyder, line 119: Could also make this more generically named `query`.

mfridman: Would it make sense to have a `ListLabelsFilter` message with all the possible filters for a given request?

```proto
ListLabelsFilter filters = 6;
```

bufdev (approved):

buf/registry/module/v1/label_service.proto
nicksnyder: In a live convo, decided that we wanted to keep the datastructure flat and instead to call it `name_query`.

buf/registry/module/v1/label_service.proto
```diff
@@ -115,6 +115,8 @@ message ListLabelsRequest {
   //
   // TODO: Add custom CEL validation to validate the status field is one of DISABLED, PASSED, APPROVED.
   repeated CommitCheckStatus commit_check_statuses = 5 [(buf.validate.field).repeated.items.enum.defined_only = true];
+  // Only return Labels with a name that contains this string.
+  string name_query = 6 [(buf.validate.field).string.max_len = 250];
```
saquibmian, line 119: Is this match case?

PR #89, opened 2024-03-09
Add ListCommits

nicksnyder:
ListCommits copies the structure from ListLabels and adds an `id_query` field similar to the `name_query` on ListLabels.

This makes it possible for a user to use a short commit hash to find the one (or multiple) actual commits that match (for typeahead search in the label selector dropdown).

### DIFF
--- a/buf/registry/module/v1/commit_service.proto
+++ b/buf/registry/module/v1/commit_service.proto
@@ -28,6 +28,10 @@ service CommitService {
   rpc GetCommits(GetCommitsRequest) returns (GetCommitsResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // List Commits for a given Module, Label, or Commit.
+  rpc ListCommits(ListCommitsRequest) returns (ListCommitsResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
 }
 
 message GetCommitsRequest {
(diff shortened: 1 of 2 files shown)


bufdev (approved):

buf/registry/module/v1/commit_service.proto
```diff
@@ -49,3 +53,51 @@ message GetCommitsResponse {
...
+  // The order to return the Commits.
+  //
+  // If not specified, defaults to ORDER_CREATE_TIME_ASC.
+  //
+  // TODO: Do we want ORDER_CREATE_TIME_ASC to be the default?
+  // TODO: We are purposefully not making the default the zero enum value, however
+  // we may want to consider this.
+  Order order = 4 [(buf.validate.field).enum.defined_only = true];
+  // Only return Commits with an id that contains this string.
+  string id_query = 5 [(buf.validate.field).string.max_len = 32];
```
saquibmian, line 93: These queries look like search queries. I was going to suggest `v1alpha1.SearchService`, but it seems like that doesn't support searching for Commits, and also doesn't support constraining results to a Module. My $0.02 is that this type of request is better implemented in Search.

Also, do we want a minimum length here?

saquibmian: I assume this is match case?

buf/registry/module/v1beta1/commit_service.proto
```diff
@@ -57,3 +61,51 @@ message GetCommitsResponse {
   // The found Commits in the same order as requested.
   repeated Commit commits = 1 [(buf.validate.field).repeated.min_items = 1];
 }
+
+message ListCommitsRequest {
```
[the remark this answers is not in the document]
nicksnyder: yeah, will add

buf/registry/module/v1/commit_service.proto
nicksnyder: We will want/need a generic search endpoint in v1, but we need to think about search holistically, and I don't want to open that can of worms yet. I think it is perfectly reasonable to (1) have List endpoints and (2) have implicitly ANDed filters on List endpoints.

> do we want a minimum length here?

Are you worried about performance? In the absence of a concrete concern I don't see a reason to artificially require a minimum length.

> I assume this is match case?

As documented, yes, but it is an interesting question I did not think explicitly about.

Commit IDs are canonically always lowercase, and people tend to type in lowercase (why push an extra button when you don't need to), so it probably doesn't matter much in this case (and being case sensitive is simpler).

That said, it makes more of a difference with labels. If I have a label called "Prod" and I search for "prod", I might be surprised if it doesn't show up.

What would you expect as a user?

saquibmian: I was more worried about non-sensical results here. If someone invokes this RPC with `id_query = 'u'`, they will get results back, but it probably doesn't make sense to fire this off before there's enough to match on. I've usually seen endpoints require min query lengths of `3`. The frontend can guard for this (and debounce), though, so it need not be in the API.

I would expect all "keys" (i.e., IDs[^0], surrogate IDs) to be case sensitive, and everything else to be case insensitive.

[^0]: I know UUIDs are case insensitive, but the fact that our IDs are UUIDs is not something that users should rely on, they are opaque strings.

nicksnyder: > The frontend can guard for this (and debounce), though, so it need not be in the API.

I agree this feels like more of a frontend consideration than something we need to enforce at the API layer.

> I would expect all "keys" (i.e., IDs[1](https://github.com/bufbuild/registry-proto/pull/89#user-content-fn-0-faa31e31f0257582b4ac7fb663baf53e), surrogate IDs) to be case sensitive, and everything else to be case insensitive.

A hypothetical search endpoint should be able to do either, but I think for API consistency we should stick to one strategy for these kinds of filters on List endpoints. If we have to pick one, picking case sensitive feels like the safest, even if it is slightly less useful for labels.

nicksnyder: After discussion, decide on case-insensitive by default: https://github.com/bufbuild/registry-proto/pull/91

PR #92, opened 2024-03-14
Change how archived labels work

nicksnyder:
Behavior changes for archived labels:
1. Archived labels can be returned by any API.
2. Archived labels are now valid ResourceRefs.
3. Add explicit filter parameter to ListLabels so it can return only unarchived labels (the default), both archived and unarchived labels (needed by the UI), or only archived labels (could be useful for the future).
4. Add UnarchiveLabels RPC so the frontend can implement an "undo" button if someone accidentally archives a label (or wants to unarchive a label for any other reason).

### DIFF
--- a/buf/registry/module/v1/label.proto
+++ b/buf/registry/module/v1/label.proto
@@ -40,7 +40,7 @@ message Label {
   // The time the Label was archived if it is currently archived.
   //
   // If this field is not set, the Label is not currently archived.
-  google.protobuf.Timestamp archived_time = 4;
+  google.protobuf.Timestamp archive_time = 4;
   // The name of the Label.
   //
   // Unique within a given Module.
--- a/buf/registry/module/v1/label.proto
+++ b/buf/registry/module/v1/label.proto
@@ -37,8 +37,10 @@ message Label {
   google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
   // The last time the Label was updated on the BSR.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
-  // Whether or not the Label is archived.
-  bool archived = 4;
+  // The time the Label was archived if it is currently archived.
+  //
+  // If this field is not set, the Label is not currently archived.
+  google.protobuf.Timestamp archived_time = 4;
   // The name of the Label.
   //
   // Unique within a given Module.
--- a/buf/registry/module/v1/label.proto
+++ b/buf/registry/module/v1/label.proto
@@ -37,10 +37,7 @@ message Label {
   google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
   // The last time the Label was updated on the BSR.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
-  // Whether or not the Label was archived.
-  //
-  // If a Label is archived, the API considers it to not exist and will not return archived
-  // Labels, except when calling LabelService.GetLabels.
+  // Whether or not the Label is archived.
   bool archived = 4;
   // The name of the Label.
   //
--- a/buf/registry/module/v1/label_service.proto
+++ b/buf/registry/module/v1/label_service.proto
@@ -54,6 +54,12 @@ service LabelService {
   rpc ArchiveLabels(ArchiveLabelsRequest) returns (ArchiveLabelsResponse) {
     option idempotency_level = IDEMPOTENT;
   }
+  // Unarchive existing Labels.
+  //
+  // This operation is atomic. Either all Labels are unarchived or an error is returned.
+  rpc UnarchiveLabels(UnarchiveLabelsRequest) returns (UnarchiveLabelsResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
 }
 
 message GetLabelsRequest {
@@ -226,3 +232,13 @@ message ArchiveLabelsRequest {
 }
 
 message ArchiveLabelsResponse {}
+
+message UnarchiveLabelsRequest {
+  // The Labels to unarchive.
+  repeated LabelRef label_refs = 1 [
+    (buf.validate.field).repeated.min_items = 1,
+    (buf.validate.field).repeated.max_items = 250
+  ];
+}
+
+message UnarchiveLabelsResponse {}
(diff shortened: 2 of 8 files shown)


buf/registry/module/v1/label_service.proto
```diff
@@ -54,6 +54,12 @@ service LabelService {
   rpc ArchiveLabels(ArchiveLabelsRequest) returns (ArchiveLabelsResponse) {
     option idempotency_level = IDEMPOTENT;
   }
+  // Unarchive existing Labels.
+  //
+  // This operation is atomic. Either all Labels are unarchived or an error is returned.
```
unmultimedio, line 59: Question that's also relevant for `ArchiveLabels` RPC: what if they're trying to unarchive something that's already unarchived? I'd say that's ok and we unarchive all the rest?

unmultimedio (approved):
lgtm

buf/registry/module/v1/label_service.proto
nicksnyder: Yes, that makes sense to me and is how I would implement this RPC.

buf/registry/module/v1/label.proto
```diff
@@ -37,10 +37,7 @@ message Label {
   google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
   // The last time the Label was updated on the BSR.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
-  // Whether or not the Label was archived.
-  //
-  // If a Label is archived, the API considers it to not exist and will not return archived
-  // Labels, except when calling LabelService.GetLabels.
```
saquibmian, line 43: This has wider-ranging consequences. Have we decided to allow users to depend on labels that are archived? Can I still consume this label via Generated SDKs?

nicksnyder: I think those are orthogonal decisions decoupled with the API we expose. The API exposes the information and clients (including the CLI) can decide what they want to do or not do with that information.

The CLI could decide to allow, warn, or error if a module is depending on an archived label. I think any choice here could be defensible, but I don't think this is an API concern. If the CLI is currently relying on archived labels not being returned at all, and that is the behavior that we want to preserve, then we will need to update the CLI to use the correct filter on ListLabels.

AFAIK generated SDKs is an implementation detail of the BSR and doesn't use the API anyway, so no matter what this API is, generated SDKs could do whatever it wanted, right?

I think it is good to keep in mind that the original intent of archiving is purely for user presentation to hide things that are no longer relevant. Whether we should prevent archived labels from being used is a step further than what we strictly need. Given we tend to want to minimize breaking builds, and given the UI we have discussed will allow you to browse an archived label, I am inclined to not prevent dependencies on archived labels (both in CLI and Generated SDKs).

Will it cause a problem if a label cycles between archived and unarchived without a new commit getting pushed?

saquibmian: The decision may be orthogonal, but you cannot change one without impacting the other. For example, how we enforce that commits can/cannot be consumed through archived labels is directly tied to API behavior:

- Before this PR, archived labels are not valid `ResourceRef`s in all but one API, so we have effectively decided that users _cannot_ consume commits through archived labels.
- With this PR, without any changes to the CLI, we are effectively taking the position that users _can_ consume commits through archived labels. 
    - In addition, if any past CLI used these APIs, we basically cannot make these changes without reversing course on our decision. This is not true today, but my point is that API changes are not in a vacuum from API consumers.

I'm not questioning the behaviour change -- that is, like you mentioned, orthogonal -- but I want to understand what the downstream impacts of this are and if any thinking has changed there. Generated SDKs is not impacted by APIs, but it's another downstream impact that I'm thinking about.

nicksnyder: I thought I reviewed all locations in the API the referenced archived labels, but I did miss some (e.g. on ResourceRef) so will take a look at those that I missed next.

nicksnyder:
Closing this for now. The current semantics of archived labels are really intended to be "Delete". One of the problems that we were trying to solve for here is making it possible for users to fix a mistake, but the current proposal creates a lot of API complexity to accomplish that. Will explore alternative ways to undo accidental deletion.

buf/registry/module/v1/label_service.proto
```diff
@@ -80,6 +86,16 @@ message ListLabelsRequest {
     // Order by create_time newest to oldest.
     ORDER_CREATE_TIME_DESC = 2;
   }
+  // The filter on the archived field of a Label.
+  enum ArchivedFilter {
```
bufdev, line 90: `ArchiveFilter`

buf/registry/module/v1/label_service.proto
```diff
@@ -80,6 +86,16 @@ message ListLabelsRequest {
...
+    ARCHIVED_FILTER_UNSPECIFIED = 0;
+    // Return only unarchived labels.
+    ARCHIVED_FILTER_ONLY_UNARCHIVED = 1;
```
bufdev, line 93: `ARCHIVED_ONLY`

buf/registry/module/v1/label_service.proto
```diff
@@ -80,6 +86,16 @@ message ListLabelsRequest {
...
+    // Return only archived labels.
+    ARCHIVED_FILTER_ONLY_ARCHIVED = 2;
```
bufdev, line 95: `ARCHIVED_ONLY`

buf/registry/module/v1/label_service.proto
```diff
@@ -80,6 +86,16 @@ message ListLabelsRequest {
...
+    // Return both archived and unarchived labels.
+    ARCHIVED_FILTER_NONE = 3;
```
bufdev, line 97: `ALL`

buf/registry/module/v1/label.proto
nicksnyder: > Have we decided to allow users to depend on labels that are archived? Can I consume this label via Generated SDKs?

The intent moving forward is that archived labels are valid ResourceRefs, so you should continue to be able to depend on them and use them in Generated SDKs.

nicksnyder (commented):
After discussing alternatives, including https://github.com/bufbuild/registry-proto/pull/94, we decided to pursue the approach proposed in this PR. That approach being:
* Archived labels can be returned by any API.
* Archived labels are now valid ResourceRefs.
* ListLabels can filter on archive state.

bufdev (approved):

PR #93, opened 2024-03-14
Change Label.archived to Label.archived_time

nicksnyder:
This is so that we can render in the UI the time a label was archived.

Stacked on top of https://github.com/bufbuild/registry-proto/pull/92 since this part is a breaking change.

### DIFF
--- a/.github/workflows/buf.yaml
+++ b/.github/workflows/buf.yaml
@@ -33,7 +33,6 @@ jobs:
     if: github.ref == 'refs/heads/main'
     needs:
       - lint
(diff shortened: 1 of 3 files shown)


.github/workflows/buf.yaml
```diff
@@ -33,7 +33,6 @@ jobs:
     if: github.ref == 'refs/heads/main'
     needs:
       - lint
-      - breaking
```
nicksnyder, line 36: This will get reverted once this change is merged into main.

PR #94, opened 2024-03-14
Delete labels instead of archiving them

nicksnyder:
The name "archive" is a bit confusing given the semantics are closer to "delete". By calling this action delete, we simplify documentation because it is implied that you will not get a deleted object back from any APIs, including GetLabels. This also means we don't need a "deleted" (or "archived") property on a Label either.

In addition to the existing behavior that a label gets automatically undeleted if/when new content is pushed to it, this change also adds and explicit undelete so it is possible to undo a deletion if it was accidental without needing to push new content.


buf/registry/module/v1/label.proto
```diff
@@ -37,11 +37,8 @@ message Label {
   google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
   // The last time the Label was updated on the BSR.
   google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
-  // Whether or not the Label was archived.
-  //
-  // If a Label is archived, the API considers it to not exist and will not return archived
-  // Labels, except when calling LabelService.GetLabels.
-  bool archived = 4;
+  reserved 4;
+  reserved "archived";
```
nicksnyder, line 41: could adjust all the field numbers if we would rather break and fix the CLI and the BSR

bufdev: We don't need to reserve anything - this API hasn't been released yet

bufdev:
From an API standpoint, I'd still make the contention that based on commonly-understood usages of the words in software, we are not deleting the labels, we are archiving them - that is actually what the action is doing. Delete implies wiping them away, at least to me.

nicksnyder:
After further discussion, going to pursue https://github.com/bufbuild/registry-proto/pull/92 instead

PR #95, opened 2024-03-15
Default to ORDER_CREATE_TIME_DESC

nicksnyder:
This resolves 2 kinds of TODOs:
1. Defaults ordering to DESC instead of ASC for create time. In the abstract, it is generally more useful to see things that have been created recently because recency is a proxy for relevance. In the frontend, we would more frequently want the recent items anyway (e.g. commits). GitHub REST API also defaults to DESC for timestamp fields. Either way this decision for a default is not a big deal, we should decide and remove the TODO.
2. It seems like we need to define the behavior of unspecified anyway so I don't see any value in explicitly setting the default value here. My proposal is to keep the default unset and just remove the TODO.

### DIFF
--- a/.github/workflows/buf.yaml
+++ b/.github/workflows/buf.yaml
@@ -33,7 +33,7 @@ jobs:
     if: github.ref == 'refs/heads/main'
     needs:
       - lint
-      - breaking
+      # - breaking
     steps:
       - uses: actions/checkout@v4
       - name: Install buf cli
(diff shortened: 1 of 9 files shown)


bufdev (commented):
If DESC is default most of the time, not: we should make the common default the 1 value, which I think means swapping DESC and ASC in most/all the enums and making the DESC value 1

PR #101, opened 2024-03-31
Convert UUIDs to TUUIDs

bufdev:
Relies on https://github.com/bufbuild/protovalidate/pull/189. Will not build until that PR is pushed through and `buf mod update` is run.

See https://github.com/bufbuild/buf/pull/2854.


nicksnyder (approved):
Pushed 6618cfd400477cbc321217f057d20b40a1951578 to update docs for ResourceRef

PR #103, opened 2024-04-25
Update Label.updated_by_user_id to be not required

nicksnyder:
Users can get deleted. This is consistent with the behavior already documented for Commit.created_by_user_id

https://github.com/bufbuild/registry-proto/blob/5a40a3fc390ac8316ca34c7776aa2f118be288b4/buf/registry/module/v1/commit.proto#L62-L65

### DIFF
--- a/buf/registry/module/v1/commit.proto
+++ b/buf/registry/module/v1/commit.proto
@@ -62,7 +62,10 @@ message Commit {
   // The id of the User that created this Commit on the BSR.
   //
   // May be empty if the User is no longer available.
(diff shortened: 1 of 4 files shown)


saquibmian:
I think you also need
```proto
```
Without this, the default value (empty string) [fails validation](https://buf.build/bufbuild/protovalidate/file/main:buf/validate/validate.proto#L3057).

We will want to update `Commit.created_by_user_id` as well.

saquibmian (approved):

bufdev (approved):

PR #105, opened 2024-04-30
[Do not merge] Prototype new GitHub workflows for push++

nicksnyder:
Creating a draft PR so it is easy to have discussions about and to inform decisions we are making with the unified GitHub action project.


.github/workflows/buf-push.yaml
```diff
@@ -0,0 +1,24 @@
...
+          version: 1.32.0-beta.1 # remove this flag when 1.32.0 is released so we go back to always tracking latest.
+                                 # todo: use .bufversion or the env variable?
+          github_token: ${{ github.token }} # is this necessary?
+      - name: buf build 
+        run: buf build --error-format github-actions
+      - name: buf format
+        run: buf format --diff --exit-code --error-format github-actions
+      - name: buf lint 
+        run: buf lint --error-format github-actions
+      - name: buf push
```
nicksnyder, line 21: as currently implemented, push doesn't happen on a branch if build, format, or lint fail

.github/workflows/buf-push.yaml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  buf:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install buf cli
+        uses: bufbuild/buf-setup-action@v1
+        with:
+          version: 1.32.0-beta.1 # remove this flag when 1.32.0 is released so we go back to always tracking latest.
+                                 # todo: use .bufversion or the env variable?
```
nicksnyder, line 13: this line could be deleted if the action already supported .bufversion, which is in the proposal, but not implemented by the current action

.github/workflows/buf-push.yaml
nicksnyder: do we want to push even if lint and build fail?

PR #106, opened 2024-04-30
[Do not merge] Prototype GitHub workflows

emcfarlane:
Continuation on #105 to help test alternative flows.

This PR syncs all pushes of branches and tags. Any delete events will archive the label. Pull requests run checks, and finally only pushes to main additionally run checks.

### DIFF
--- a/.github/workflows/buf-delete.yaml
+++ b/.github/workflows/buf-delete.yaml
@@ -1,0 +1,16 @@
+name: Buf delete
+on:
+  delete
+permissions: read-all
+jobs:
+  buf-push:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Buf setup
+        uses: bufbuild/buf-setup-action@v1
+        with:
+          version: 1.32.0-beta.1
+      - run: buf --version
+      - name: Buf delete
+        run: echo "Action not supported for buf"
--- a/.github/workflows/buf-pr.yaml
+++ b/.github/workflows/buf-pr.yaml
@@ -1,0 +1,31 @@
+name: Buf PR
+on:
+  pull_request:
+    types: [opened,synchronize,reopened,labeled,unlabeled]
+permissions: read-all
+jobs:
+  buf-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Buf setup
+        uses: bufbuild/buf-setup-action@v1
+        with:
+          version: 1.32.0-beta.1
+      - run: buf --version
+      - name: Buf build
+        run: buf build . --error-format github-actions
+      - name: Buf format
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'Buf Skip Format') }}
+        run: buf format . --diff --error-format github-actions --exit-code
+      - name: Buf lint
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'Buf Skip Lint') }}
+        run: buf lint . --error-format github-actions
+      - name: Checkout base
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: .base
+      - name: Buf breaking
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'Buf Skip Breaking') }}
+        run: buf breaking . --error-format github-actions --against .base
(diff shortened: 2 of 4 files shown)


.github/workflows/buf-pr.yaml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Buf setup
+        uses: bufbuild/buf-setup-action@v1
+        with:
+          version: 1.32.0-beta.1
+      - run: buf --version
+      - name: Buf build
+        run: buf build . --error-format github-actions
```
nicksnyder, line 17: I don't think the dot is necessary? I would clean this up everywhere
```suggestion
        run: buf build --error-format github-actions
```

.github/workflows/buf-pr.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Buf format
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'Buf Skip Format') }}
```
nicksnyder, line 19: Thoughts on `Buf Skip Format` vs `Skip Buf Format` vs `Skip buf format` vs `skip buf format`? (same question for other instances)

.github/workflows/buf-push-checks.yaml
```diff
@@ -0,0 +1,32 @@
+name: Buf push checks
+on:
+  push:
+    branches:
+      - main
```
nicksnyder, line 5: I think it will be better to have a single workflow file for all pushes.

Needing to hard code `main` here isn't great. In a unified workflow we can use `${{ github.event.repository.default_branch }}` to detect if we are on the right branch and behave accordingly.

.github/workflows/buf-delete.yaml
```diff
@@ -0,0 +1,16 @@
+name: Buf delete
```
nicksnyder, line 1: I assume this is deleting a label (would indicate in the name)

.github/workflows/buf-delete.yaml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Buf setup
+        uses: bufbuild/buf-setup-action@v1
+        with:
+          version: 1.32.0-beta.1
+      - run: buf --version
+      - name: Buf delete
+        run: echo "Action not supported for buf"
```
nicksnyder, line 16: I would add a comment for the command that we want to run when it is implemented (presumably `buf beta registry label archive`)

.github/workflows/buf-delete.yaml
emcfarlane: The `buf-pr` is for a pull request event, so this `buf-delete` is for a delete event but the action is to archive the label.

.github/workflows/buf-pr.yaml
emcfarlane: I originally thought Title Case was more consistent with most labels, but lower case would be my preference. Also I think it works better for checks within the merge commit message.

.github/workflows/buf-push-checks.yaml
emcfarlane: Starter workflows have `$default-branch` templating: https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization#creating-a-starter-workflow
And other workflow files will specify the default as part of their config. I don't think it's possible to access the github event as part of the specification of what events it listens on as the definition would be circular.

.github/workflows/buf-delete.yaml
nicksnyder: yeah makes sense, I figured that out and thought I deleted this coment

.github/workflows/buf-pr.yaml
nicksnyder: going to continue this conversation on https://github.com/bufbuild/buf-action/pull/9

.github/workflows/buf-push-checks.yaml
nicksnyder: Is buf-action using starter workflows?

> I don't think it's possible to access the github event as part of the specification of what events it listens on as the definition would be circular.

No you can't, but you can use `if:` statements for the jobs or steps.
https://stackoverflow.com/questions/64781462/github-actions-default-branch-variable

emcfarlane: Not currently, but it's something we could provide in addition to reusable workflows. I could see the use case where a user  wants to edit the pre-defined workflow and this would provide an eject mechanism.

PR #108, opened 2024-05-07
Fix documentation and validation on `UpdateModulesRequest.default_label_name`

saquibmian:
If present, it must be non-empty and may not point to an archived label.

### DIFF
--- a/buf/registry/module/v1/module_service.proto
+++ b/buf/registry/module/v1/module_service.proto
@@ -168,8 +168,11 @@ message UpdateModulesRequest {
     //
     // This Label may not yet exist.
     //
-    // This may point to an archived Label.
-    optional string default_label_name = 7 [(buf.validate.field).string.max_len = 250];
+    // This may not point to an archived Label.
+    optional string default_label_name = 7 [
+      (buf.validate.field).string.min_len = 1,
+      (buf.validate.field).string.max_len = 250
+    ];
   }
   // The Modules to update.
   repeated Value values = 1 [
--- a/buf/registry/module/v1beta1/module_service.proto
+++ b/buf/registry/module/v1beta1/module_service.proto
@@ -168,8 +168,11 @@ message UpdateModulesRequest {
     //
     // This Label may not yet exist.
     //
-    // This may point to an archived Label.
-    optional string default_label_name = 7 [(buf.validate.field).string.max_len = 250];
+    // This may not point to an archived Label.
+    optional string default_label_name = 7 [
+      (buf.validate.field).string.min_len = 1,
+      (buf.validate.field).string.max_len = 250
+    ];
   }
   // The Modules to update.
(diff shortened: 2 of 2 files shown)


buf/registry/module/v1/module_service.proto
```diff
@@ -168,8 +168,11 @@ message UpdateModulesRequest {
     //
     // This Label may not yet exist.
     //
-    // This may point to an archived Label.
-    optional string default_label_name = 7 [(buf.validate.field).string.max_len = 250];
+    // This may not point to an archived Label.
+    optional string default_label_name = 7 [
+      (buf.validate.field).string.min_len = 1,
```
nicksnyder, line 173: We need `(buf.validate.field).ignore = IGNORE_IF_UNPOPULATED` so that you are allowed to not specify a default label name, right?

saquibmian: The default `IGNORE_UNSPECIFIED` will skip validation if it's unset: https://github.com/bufbuild/protovalidate/blob/d52c9e35c96a0c5191aa1b1d24291958037ed9c8/proto/protovalidate/buf/validate/validate.proto#L207

Usually `IGNORE_UNPOPULATED` is specified with a `field.required`, because that one [requires that it's set](https://github.com/bufbuild/protovalidate/blob/d52c9e35c96a0c5191aa1b1d24291958037ed9c8/proto/protovalidate/buf/validate/validate.proto#L137-L143).

FWIW, I tested with with the following test:
```go
msg := &modulev1.UpdateModulesRequest{
    Values: []*modulev1.UpdateModulesRequest_Value{
        {
            ModuleRef: &modulev1.ModuleRef{
                Value: &modulev1.ModuleRef_Id{
                    Id: "1b785b565b964fc8bbe78fc8fa560654",
                },
            },
        },
    },
}
protoValidator, err := protovalidate.New()
require.NoError(t, err)
require.NoError(t, protoValidator.Validate(msg))
```
As soon as I set a value, even if it's empty, the validation rules apply:
```go
var defaultLabel string
msg := &modulev1.UpdateModulesRequest{
    Values: []*modulev1.UpdateModulesRequest_Value{
        {
            ModuleRef: &modulev1.ModuleRef{
                Value: &modulev1.ModuleRef_Id{
                    Id: "1b785b565b964fc8bbe78fc8fa560654",
                },
            },
            DefaultLabelName: &defaultLabel,
        },
    },
}
protoValidator, err := protovalidate.New()
require.NoError(t, err)
require.NoError(t, protoValidator.Validate(msg))
```
fails with
> - values[0].default_label_name: value length must be at least 1 characters [string.min_len]

nicksnyder: got it, thanks for checking!

nicksnyder (approved):

PR #115, opened 2024-06-03
Use buf-action in CI

emcfarlane:
This updates CI to use the new `buf-action` replacing `buf-setup`, `buf-lint`, `buf-breaking` and `buf-push` unifying these tasks as one. The new behaviour is as follows:
- On pull requests: run build, lint, format and breaking change detection checks.
- On push: build and then push the changes of the latest commit with the label of the current branch (e.g. main).
- On delete: archive the label associated with the delete (e.g. the branch name of the PR).

Change in behaviour:
- Buf version has been set to `1.32.2`, the current latest release.
- Checks no longer block pushes of commits to the BSR.
- Formatting checks have been added to `buf-ci.yaml`.
- Go version has been bumped to the latest minor release, for maintenance reasons only. 
- Use the setup-go cache for better caching of build artifacts.


.github/workflows/buf-ci.yaml
```diff
@@ -0,0 +1,18 @@
+name: buf
```
emcfarlane, line 1: File renamed for improved support for IDEs as `buf.yaml` can check it implemented a buf configuration file.

.github/workflows/buf-ci.yaml
```diff
@@ -0,0 +1,18 @@
...
+on:
+  push:
+    branches:
+      - main
```
emcfarlane, line 5: Trigger push only on main, same as current.

.github/workflows/ci.yaml
```diff
@@ -12,12 +12,13 @@ jobs:
       - name: install
         uses: actions/setup-go@v5
         with:
-          go-version: 1.21.x
+          go-version: 1.22.x
+          cache: false
```
emcfarlane, line 16: Cache isn't used as no `go.sum` present, but explicitly disable to avoid warning.

.bufversion
```diff
@@ -0,0 +1 @@
+1.32.2
```
bufdev, line 1: Not approving this as our pattern - this would imply to me as a third-party that we have support for this out of the box with our toolchain, which we do not

emcfarlane: We support this behaviour through [bufisk](https://github.com/bufbuild/bufisk). If not using the `.bufversion` file we can use the env var `BUF_VERSION`. This could be set in the main actions environment or in each workflow file.

bufdev: Bufisk was an experiment that I wrote up in one day and never finished, we never got to stable, and don't advertise. It effectively isn't officially supported. Outside of that, there is no support for `.bufversion`.

.github/workflows/buf-ci.yaml
nicksnyder: I don't think we have a good reason to deviate from what we recommend to others.

.github/workflows/buf-ci.yaml
```diff
@@ -0,0 +1,20 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+permissions:
+  contents: read
+  pull-requests: write
+env:
+  BUF_VERSION: 1.32.2
```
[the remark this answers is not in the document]
emcfarlane: Env to allow setting once and sharing between tools, but theres no use case here. Reverted and set as `version:` input to the action.

.github/workflows/buf-ci.yaml
emcfarlane: Enabled sync: push and delete for branches.

nicksnyder:
PR description needs an update based on changes from review comments (e.g. `.bufversion`, pushing on any branch)

.github/workflows/ci.yaml
nicksnyder: can/should this be a comment in the yaml itself?

emcfarlane: I've changed this to use `cache-dependent-path` in the setup-go action which looks to be much better at speeding up the workflow (I think from caching build artifacts). For example:
- cache hit using setup-go: https://github.com/bufbuild/registry-proto/actions/runs/9357564225/job/25757623058
- cache hit using the custom cache step: https://github.com/bufbuild/registry-proto/actions/runs/9357179796/job/25756335481

Results: 6s for setup-go vs 41s for toolkit cache.

nicksnyder (approved):
This is what I would expect, but should wait for @bufdev to approve too.

bufdev (approved):

PR #117, opened 2024-06-03
Add protovalidate rules for RPCs where UNSPECIFIED enum values are not accepted

doriable:
For some RPCs, the `_UNSPECIFIED` value is not accepted and handled.
This situation came up in a circumstance when testing, putting an
`_UNSPECIFIED` enum value for an RPC that does not handle this creates
an error in our monitoring, when it should return an `InvalidArgument` through
our validation rules to the user (without alerting us through monitoring).

I went through the RPCs and enums and looked for cases where we do not
accept `_UNSPECIFIED` and added protovalidate rules.


buf/registry/module/v1/download_service.proto
```diff
@@ -48,7 +48,10 @@ message DownloadRequest {
     // Specific file types to request.
     //
     // If not set, all file types are returned.
-    repeated FileType file_types = 2 [(buf.validate.field).repeated.items.enum.defined_only = true];
+    repeated FileType file_types = 2 [
+      (buf.validate.field).repeated.items.enum.defined_only = true,
+      (buf.validate.field).repeated.items.enum.not_in = 0
```
pkwarren, line 53: Should we add the unique constraint here too?

buf/registry/module/v1beta1/download_service.proto
```diff
@@ -49,7 +49,10 @@ message DownloadRequest {
     // Specific file types to request.
     //
     // If not set, all file types are returned.
-    repeated FileType file_types = 2 [(buf.validate.field).repeated.items.enum.defined_only = true];
+    repeated FileType file_types = 2 [
+      (buf.validate.field).repeated.items.enum.defined_only = true,
+      (buf.validate.field).repeated.items.enum.not_in = 0
```
pkwarren, line 54: Unique constraint?

buf/registry/module/v1/download_service.proto
pkwarren, line 53:
```suggestion
      (buf.validate.field).repeated.items.enum.not_in = [0]
```

buf/registry/owner/v1/user_service.proto
```diff
@@ -97,12 +97,14 @@ message ListUsersRequest {
...
-    (buf.validate.field).repeated.items.enum.defined_only = true
+    (buf.validate.field).repeated.items.enum.defined_only = true,
+    (buf.validate.field).repeated.items.enum.not_in = 0
   ];
   // Only return Users of these states.
   repeated UserState has_states = 6 [
     (buf.validate.field).repeated.unique = true,
-    (buf.validate.field).repeated.items.enum.defined_only = true
+    (buf.validate.field).repeated.items.enum.defined_only = true,
+    (buf.validate.field).repeated.items.enum.not_in = 0
```
pkwarren, line 107: Makes sense to add unique to these as well.

buf/registry/module/v1/download_service.proto
rodaine: For repeated fields being set _not_ in a message literal, you use this singular format (and repeat the same line to add more), and do not wrap it in `[]`

doriable: Agreed, added here and to `v1beta1`

buf/registry/owner/v1/user_service.proto
```diff
@@ -128,7 +130,10 @@ message CreateUsersRequest {
     // The type of the User.
     //
     // If not set, the default USER_TYPE_STANDARD is used.
-    UserType type = 2 [(buf.validate.field).enum.defined_only = true];
+    UserType type = 2 [
+      (buf.validate.field).enum.defined_only = true,
+      (buf.validate.field).enum.not_in = 0
```
bufdev, line 135: Doesn't this effectively make it required, given that the field is by default the zero value?

rodaine: A bit more restrictive since you could pass 999 with just required, but defined_only limits it to the defined values.

nicksnyder: I think the observation here is that if we add this validation, the field is changing from effectively optional (there is a defined default behavior if you pass the zero value) to required (you must provide a value that is not the zero value). 

Does not setting a value for a non-optional enum send the `0` value over the wire?

doriable: Double checked everything, I think we do not want to restrict `not_in: 0` here since we do actually want to allow `_UNSPECIFIED` to be passed and we use the fallback condition outlined. Going to remove this.

buf/registry/owner/v1/user_service.proto
```diff
@@ -97,12 +97,14 @@ message ListUsersRequest {
   // Only return Users of these types.
   repeated UserType has_types = 5 [
     (buf.validate.field).repeated.unique = true,
-    (buf.validate.field).repeated.items.enum.defined_only = true
+    (buf.validate.field).repeated.items.enum.defined_only = true,
+    (buf.validate.field).repeated.items.enum.not_in = 0
```
doriable, line 101: Question for this: when `UNSPECIFIED` is passed, we are currently falling back to the same type as `STANDARD`, however that functionally bypasses the uniqueness check (e.g. if they have both `STANDARD` and `UNSPECIFIED`). Should we continue disallow this?

rodaine: For a repeated field of enums that's intended to be unique as well, it makes sense to me to exclude UNSPECIFIED as a valid option. In this case, a user cannot really have an UNSPECIFIED type.

buf/registry/owner/v1/user_service.proto
bufdev: I avoided adding unique constraints as it was unclear what hit it'd have on perf as n approaches infinity - perhaps not a practical concern, but I'm not sure about them

buf/registry/owner/v1/user_service.proto
bufdev: Doesn't this call into question the use of not_in altogether?

I had not set it because of the requiredness it enforces.

rodaine: For a singular enum field where unspecified is valid, you don't want it. Elsewhere for repeated enum fields, it doesn't make sense to have unspecified as a valid option, so not_in is appropriate.

buf/registry/module/v1/module_service.proto
```diff
@@ -118,7 +118,8 @@ message CreateModulesRequest {
     // The module's visibility.
     ModuleVisibility visibility = 3 [
       (buf.validate.field).required = true,
-      (buf.validate.field).enum.defined_only = true
+      (buf.validate.field).enum.defined_only = true,
+      (buf.validate.field).enum.not_in = 0
```
nicksnyder, line 122: Was chatting with @rodaine about this w.r.t. my PR, and this not_in is redundant with required. Using required is probably better UX so propose removing this here.
```suggestion
```

buf/registry/module/v1beta1/module_service.proto
```diff
@@ -118,7 +118,8 @@ message CreateModulesRequest {
     // The module's visibility.
     ModuleVisibility visibility = 3 [
       (buf.validate.field).required = true,
-      (buf.validate.field).enum.defined_only = true
+      (buf.validate.field).enum.defined_only = true,
+      (buf.validate.field).enum.not_in = 0
```
nicksnyder, line 122: ditto
```suggestion
```

buf/registry/owner/v1/user_service.proto
doriable: Excellent, I'll keep this restriction in that case. Thank you!

pkwarren (approved):
Changes LGTM. I think adding these on just `optional` and `repeated` is a good first step and will resolve the original issue we saw.

rodaine (approved):

nicksnyder (approved):

buf/registry/owner/v1/user_service.proto
rodaine: The unique check is O(n) with fail-fast behavior (at least in the go impl). Basically, shoves the values into a `map[T]struct{}` which does mean there's some allocation cost. Unless you've got a massive amount of values, though it's fairly negligible.

bufdev (approved):

PR #119, opened 2024-06-06
Add ability to configure base ModuleRole for Organizations

nicksnyder:
We want to update the BSR web app to be able to configure the base role at creation time but the current v1 APIs don't support setting/changing this, so the current web UI is implemented by using v1 RPC to create and organization, and then an alpha API to immediately set the base role. It would be ideal if we could update the v1 API to support setting the base role at creation time.

This PR is to start a discussion about if, and how, we want to represent roles in the organization API. It is ok if the conclusion is we want to defer thinking about this until we think about roles more holistically.

### DIFF
--- a/buf/registry/owner/v1/organization.proto
+++ b/buf/registry/owner/v1/organization.proto
@@ -90,7 +90,7 @@ enum ModuleRole {
   MODULE_ROLE_WRITE = 3;
   // A role that has administrative access to the module.
   MODULE_ROLE_ADMIN = 4;
-  // A role that is the owner of the module.
+  // A role that has owner access to the module.
   MODULE_ROLE_OWNER = 5;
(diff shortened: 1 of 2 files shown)


buf/registry/owner/v1/organization.proto
```diff
@@ -72,6 +79,21 @@ enum OrganizationVerificationStatus {
   ORGANIZATION_VERIFICATION_STATUS_OFFICIAL = 3;
 }
 
+// An access role to a module.
+enum ModuleRole {
```
nicksnyder, line 83: Should this type belong in `buf.registry.module` instead of `buf.registry.organization`?

nicksnyder:
Going to close this since it is stale.

PR #122, opened 2024-07-05
Add deprecation message to v1 and v1beta1 module

oliversun9:
This PR adds deprecation message to v1 and v1beta1 module, so that the CLI command that deprecates a module can use v1 API.

### DIFF
--- a/buf/registry/module/v1/module.proto
+++ b/buf/registry/module/v1/module.proto
@@ -76,6 +76,15 @@ message Module {
     (buf.validate.field).required = true,
     (buf.validate.field).string.max_len = 250
   ];
+  // The message of deprecating the module. Typically a message redirecting readers
+  // to the module that should be used instead.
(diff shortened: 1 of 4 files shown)


oliversun9:
This is also one of the rare cases where `buf lint` fails, a fix should be up soon.

buf/registry/module/v1/module.proto
```diff
@@ -76,6 +76,15 @@ message Module {
     (buf.validate.field).required = true,
     (buf.validate.field).string.max_len = 250
   ];
+  // The message of deprecating the module. Typically a message redirecting readers
+  // to the module that should be used instead.
+  string deprecation_message = 11 [(buf.validate.field).string.max_len = 255];
```
bufdev, line 81: We did not bring this over on purpose, thus should be dropped.

bufdev (changes requested):

PR #125, opened 2024-07-10
Fix code owner rule that got overwritten

oliversun9:
Currently, the owner role on the first line is overwritten the second line, this PR fixes that.

From [GitHub's docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file):

> Order is important; the last matching pattern takes the most precedence.


nicksnyder (approved):
Thank you!

PR #126, opened 2024-07-12
Add initial specification for plugin support

smallsamantha:
This PR is a work in progress, open questions need to be addressed before full review.

- [ ] How should a Plugin indicate it supports Generated SDKs? 
- [ ] How should a Plugin indicate it is verified/managed? 
- [ ] How should a Plugin [and Module] represent its license?
- [ ] Should a filter on deprecated Plugins be provided on ListPlugins?
- [ ] Should a filter on Generated SDK support be provided on ListPlugins?
- [ ] Should an Update endpoint be provided?
- [ ] How should revision be validated?
- [ ] How should semver version be validated?
- [ ] How should digest be validated?
- [ ] Should it be specified that Modules/Plugins share a namespace (so names across both types must be unique per owner)?
- [ ] How should individual, language-specific Plugin configuration be represented?

### DIFF
--- a/buf/registry/plugin/v1/plugin.proto
+++ b/buf/registry/plugin/v1/plugin.proto
@@ -101,8 +101,8 @@ message PluginRef {
 
   // The fully-qualified name of a Plugin within a BSR instance.
   //
-  // A Name uniquely identifies a Plugin.
-  // This is used for requests when a caller only has the name of the Plugin and not the ID.
+  // In combination with its version and revision, a Name uniquely identifies a Plugin.
+  // This is used for requests when a caller only has the name and version components of the Plugin and not the ID.
   message Name {
     // The name of the owner of the Plugin, either a User or Organization.
     string owner = 1 [
--- a/buf/registry/plugin/v1/plugin.proto
+++ b/buf/registry/plugin/v1/plugin.proto
@@ -1,0 +1,130 @@
+// Copyright 2023-2024 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.registry.plugin.v1;
+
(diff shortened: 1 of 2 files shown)


buf/registry/plugin/v1/plugin_service.proto
```diff
@@ -0,0 +1,156 @@
...
+package buf.registry.plugin.v1;
+
+import "buf/registry/owner/v1/owner.proto";
+import "buf/registry/plugin/v1/plugin.proto";
+import "buf/validate/validate.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/plugin/v1";
+
+// Operate on Plugins.
+service PluginService {
```
mfridman, line 26: Not sure if we're planning to bring in the [ResolveService](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.ResolveService), but it'd be nice if we stubbed out a single RPC to resolve Generated SDK version. Something like

```
rpc ResolveGeneratedSDKVersion
```

Basically folding all the `Get{Go,Swift,Maven,NPM,...}Version` into a single RPC. You could argue that this fits nicely with the "PluginService".

saquibmian: Ideally there's also an RPC to get a registry-specific package ID. Could be the same RPC.

buf/registry/plugin/v1/plugin.proto
```diff
@@ -0,0 +1,130 @@
...
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.tuuid = true
+  ];
+  // The time the Plugin was created on the BSR.
+  google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
+  // The last time the Plugin was updated on the BSR.
+  google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
+  // The name of the Plugin.
+  //
+  // Unique within a given User or Organization.
```
saquibmian, line 40: Does this imply that each version/revision of the plugin has a different name?

buf/registry/plugin/v1/plugin.proto
```diff
@@ -0,0 +1,130 @@
...
+    option (buf.validate.oneof).required = true;
+    // The id of the Plugin.
+    string id = 1 [(buf.validate.field).string.tuuid = true];
+    // The fully-qualified name of the Plugin.
+    Name name = 2;
+  }
+  // Semver-formatted release version of the Plugin.
+  string version = 3;
+  // The revision of the Plugin. This represents the BSR image version of the Plugin.
+  uint32 revision = 4;
```
saquibmian, line 129: Should these be under `Name`? Why would I want to specify these if I have the plugin ID?

saquibmian (commented):
Drive by

PR #129, opened 2024-08-02
Add error details for Upload and GetCommits

oliversun9:
This adds error details for `Upload` and `GetCommits`. This change allows servers (at least the BSR) to return structured errors when multiple commit IDs or resource refs are not found (could be extended later to account for other types of errors).

### DIFF
--- a/buf/registry/module/v1/commit_service.proto
+++ b/buf/registry/module/v1/commit_service.proto
@@ -54,6 +54,13 @@ message GetCommitsResponse {
   repeated Commit commits = 1 [(buf.validate.field).repeated.min_items = 1];
 }
 
+// GetCommitsErrorDetails is the error details for the GetCommits RPC.
+message GetCommitsErrorDetails {
+  // This contains the resource_refs that the server is unable to find. This is populated
+  // when a non-found error is returned by the server.
+  repeated ResourceRef not_found_resource_refs = 1;
+}
+
 message ListCommitsRequest {
   // The list order.
   enum Order {
--- a/buf/registry/module/v1/upload_service.proto
+++ b/buf/registry/module/v1/upload_service.proto
@@ -83,3 +83,10 @@ message UploadResponse {
   // If nothing changed for a given reference, the existing Commit will be returned.
   repeated Commit commits = 1 [(buf.validate.field).repeated.min_items = 1];
 }
+
+// UploadErrorDetails is the error details for the Upload RPC.
+message UploadErrorDetails {
+  // This contains the dep_commit_ids that the server is unable to find. This is populated
+  // when a non-found error is returned by the server.
+  repeated string not_found_dep_commit_ids = 1;
+}
--- a/buf/registry/module/v1beta1/upload_service.proto
+++ b/buf/registry/module/v1beta1/upload_service.proto
@@ -98,3 +98,19 @@ message UploadResponse {
   // If nothing changed for a given reference, the existing Commit will be returned.
   repeated Commit commits = 1 [(buf.validate.field).repeated.min_items = 1];
 }
+
+// UploadErrorDetails is the error details for the Upload RPC.
+message UploadErrorDetails {
+  message DepRef {
+    // The commit_id of the dependency.
+    string commit_id = 1 [
+      (buf.validate.field).required = true,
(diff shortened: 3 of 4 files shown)


buf/registry/module/v1/commit_service.proto
```diff
@@ -54,6 +54,13 @@ message GetCommitsResponse {
   repeated Commit commits = 1 [(buf.validate.field).repeated.min_items = 1];
 }
 
+// GetCommitsErrorDetails is the error details for the GetCommits RPC.
+message GetCommitsErrorDetails {
+  // This contains the resource_refs that the server is unable to find. This is populated
+  // when a non-found error is returned by the server.
+  repeated ResourceRef not_found_resource_refs = 1;
```
oliversun9, line 61: `ResourceRef` is `request_only`, should we expand `request_only` to allow error details?

buf/registry/module/v1beta1/upload_service.proto
```diff
@@ -98,3 +98,19 @@ message UploadResponse {
...
+    string commit_id = 1 [
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.tuuid = true
+    ];
+    // The registry hostname of the dependency.
+    string registry = 2 [(buf.validate.field).required = true];
+  }
+  // This contains the dep_refs that the server is unable to find. This is populated
+  // when a non-found error is returned by the server.
+  repeated string not_found_dep_refs = 1;
```
bufdev, line 115: Is this supposed to be `repeated DepRef`?

oliversun9: Yes, fixed in [da730f2](https://github.com/bufbuild/registry-proto/pull/129/commits/da730f285913e9b777a7653d11e29678711b12b7)

buf/registry/module/v1/upload_service.proto
```diff
@@ -83,3 +83,10 @@ message UploadResponse {
   // If nothing changed for a given reference, the existing Commit will be returned.
   repeated Commit commits = 1 [(buf.validate.field).repeated.min_items = 1];
 }
+
+// UploadErrorDetails is the error details for the Upload RPC.
+message UploadErrorDetails {
+  // This contains the dep_commit_ids that the server is unable to find. This is populated
+  // when a non-found error is returned by the server.
+  repeated string not_found_dep_commit_ids = 1;
+}
```
saquibmian, line 92: I don't have much experience with error details, but I'm wondering if we should structure these RPC-specific error details a little more like a "giant oneof". Something like this:
```proto
message UploadErrorDetails {
  message ModuleNotFoundErrorDetail {
    ModuleRef module_ref = 1;
    int32 content_index = 2;
  }
  message DepCommitNotFoundErrorDetail {
    string commit_id = 1;
  }
  message CycleDetectedErrorDetail {
    repeated string commit_ids = 1;
  }
  message UnknownFileTypeErrorDetail {
    int32 content_index = 1;
    string file_path = 2;
  }
  oneof value {
    ModuleNotFoundErrorDetail module_not_found = 1;
    DepCommitNotFoundErrorDetail dep_commit_not_found = 2;
    CycleDetectedErrorDetail cycle_detected = 3;
    UnknownFileTypeErrorDetail unknown_file_type = 4;
  }
}
```
This would enable clients to handle _cases_ of errors instead of conditionally checking if a field/set of fields are set to determine the semantics of the error. It's also very obvious what kinds of structured errors this RPC could return.

In the future, we could also define common error details that can compose nicely with this pattern, but that gets into plurality/etc. that we don't need to consider for now... just thinking out loud here.

oliversun9: I think using oneof is definitely nice. However, will there ever be a need to return both `ModuleNotFound` and `DepCommitNotFound`? If so, I think the oneof could be categorized by error code instead.

As an example of a common error detail, we could also add [`buf.validate.Violations`](https://github.com/bufbuild/protovalidate/blob/176e0a2f0fc39207e659cf765a805f67baebcb61/proto/protovalidate/buf/validate/expression.proto#L60) to the oneof in the future.

oliversun9:
Discussed offline with @smallsamantha, closing this for now.

PR #132, opened 2024-09-25
Add initial Plugin specification

mfridman:
This PR adds a new Plugin entity type. It is modelled like the Module entity.

A plugin and a module both have immutable commits and a label history. Which means they can be similarly referenced within configuration files by a label or an explicit commit.

Note, given the similarities between commits and labels, there's an alternative proposal by @doriable in https://github.com/bufbuild/registry-proto/pull/132#issuecomment-2407967961 to avoid duplicating common definitions.


doriable:
At the top-level, it seems like we are making carbon copies of definitions for resources, commits, and labels here, which makes sense from the perspective that these are shared concepts. But because of that, I also think rather than making copies, we should actually be defining our APIs to reflect the fact that they are shared. My thoughts are something that looks like:

```
(main %=)🐱~/src/buf/registry-proto @ $ tree .
.
├── buf
│   └── registry
│       ├── resource
│       │   ├── module
│       │   │   └── v1
│       │   │       ├── file.proto
│       │   │       ├── graph.proto
│       │   │       ├── graph_service.proto
│       │   │       ├── module.proto
│       │   │       └── module_service.proto
│       │   ├── plugin
│       │   │   └── v1
│       │   │       ├── code_generation.proto
│       │   │       ├── collection.proto
│       │   │       ├── plugin.proto
│       │   │       └── plugin_service.proto
│       │   └── v1
│       │       ├── commit.proto
│       │       ├── commit_service.proto
│       │       ├── digest.proto
│       │       ├── download_service.proto
│       │       ├── label.proto
│       │       ├── label_service.proto
│       │       ├── resource.proto
│       │       ├── resource_service.proto
│       │       └── upload_service.proto
│       ├── module (same as current)
│       ├── owner (same as current)
│       └── priv (same as current)
├── buf.lock
└── buf.yaml
```

Or alternatively, a slightly less nested structure:

```
(main %=)🐱~/src/buf/registry-proto @ $ tree .
.
├── buf
│   └── registry
│       ├── module
│       │   ├── v1 (same as current)
│       │   ├── v1beta1 (same as current)
│       │   └── v2
│       │       ├── file.proto
│       │       ├── graph.proto
│       │       ├── graph_service.proto
│       │       ├── module.proto
│       │       └── module_service.proto
│       ├── owner (same as current)
│       ├── plugin
│       │   └── v1
│       │       ├── code_generation.proto
│       │       ├── collection.proto
│       │       ├── plugin.proto
│       │       └── plugin_service.proto
│       ├── priv (same as current)
│       └── resource
│           └── v1
│               ├── commit.proto
│               ├── commit_service.proto
│               ├── digest.proto
│               ├── download_service.proto
│               ├── label.proto
│               ├── label_service.proto
│               ├── resource.proto
│               ├── resource_service.proto
│               └── upload_service.proto
├── buf.lock
└── buf.yaml
```

In either case though, I think the API should reflect what the shared entities are. This proposal would also come with the following changes:
- This would add `P1` as a new digest type to the overall digests, and then I think I would want to use protovalidate rules to enforce `B*` digests for module and `P1` digest for plugins.
- Download/upload contents would now be one-ofs of module content, check plugin content, code generation content
- Resource would now be one-of the following: `Module`, `Plugin`, `Label`, or `Commit`

I am open to different names other than "resource", this just felt consistent with what we have right now. But basically, I think the premise of this is that if there were any foundational changes to either labels and/or commits, right now it's _implicit_ that you need to update both modules and plugins, whereas this would make it explicit.

buf/registry/plugin/v1beta1/code_generation_config.proto
```diff
@@ -0,0 +1,333 @@
...
+syntax = "proto3";
+
+package buf.registry.plugin.v1beta1;
+
+import "buf/validate/validate.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/plugin/v1beta1";
+
+// The configuration for code generation.
+message CodeGenerationConfig {
```
emcfarlane, line 24: CodeGenerationConfig could be live in the [bufplugin](https://github.com/bufbuild/bufplugin/tree/main) repository as generation config required for the languages.

buf/registry/plugin/v1beta1/plugin.proto
```diff
@@ -0,0 +1,142 @@
...
+import "buf/registry/plugin/v1beta1/collection.proto";
+import "buf/registry/priv/extension/v1beta1/extension.proto";
+import "buf/validate/validate.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/plugin/v1beta1";
+
+message Plugin {
+  option (buf.registry.priv.extension.v1beta1.message).response_only = true;
+
```
emcfarlane, line 28: We need default label to resolve resources to a commit.

buf/registry/plugin/v1beta1/commit.proto
```diff
@@ -0,0 +1,79 @@
...
+    (buf.validate.field).string.tuuid = true
+  ];
+  // The id of the Plugin that the Commit is associated with.
+  string plugin_id = 4 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.tuuid = true
+  ];
+  // The Digest of the Commit's contents.
+  //
+  // TODO(mf): if we have both docker and wasm modules, is this only returning the p1 digest, or
```
mfridman, line 58: @emcfarlane what do you think?

buf/registry/plugin/v1beta1/plugin_service.proto
```diff
@@ -0,0 +1,229 @@
...
+    // TODO(mf): do we want this at the Plugin or the PluginCommit.
+    //
+    // I'm inclined to leave this on the Plugin. I think the most common use case is a plugin
+    // repository moving from github.com/foo/bar to github.com/new/repo and that event should apply
+    // to all commits since the project as a whole is moving, and not a subset of commits.
+    //
+    // For example, connect used to be under bufbuild but moved to a new organization, when looking
+    // at the Go docs there is a single source of truth for the project despite having a different
+    // home.
+    string source_url = 5 [
```
mfridman, line 161: IMO the `source_url` should be a property of the plugin entity and not the commit, which means a user can update this via the BSR and have it applied to _all_ commits.

> I think the most common use case is a plugin repository moving from github.com/foo/bar to
> github.com/new/repo and that **event should apply to all commits since the project as a whole is
> moving, and not a subset of commits.**
> 
> For example, connect used to be under bufbuild but moved to a new organization, when looking at
> the Go docs there is a single source of truth for the project despite having a different home.

There's a `url` in the PluginInfo, but we should ignore this value in favor of whatever is set on the BSR.

https://buf.build/bufbuild/bufplugin/docs/main:buf.plugin.info.v1#buf.plugin.info.v1.PluginInfo

buf/registry/plugin/v1beta1/commit.proto
emcfarlane: I think that makes sense.

mfridman: Updated in fd48e8f6

mfridman: Although we're not allowing users to download docker images, I think being able to retrieve the Plugin as a whole and having the digest is useful, that way you know whether you can request a commit that'll ultimately lead to a wasm module.

buf/registry/plugin/v1beta1/plugin.proto
mfridman: I think the default label should be `main`, similar to modules. I don't think this needs to change?

emcfarlane: Yes, doesn't need to change if the default label cannot be altered.

buf/registry/plugin/v1beta1/plugin_service.proto
```diff
@@ -0,0 +1,217 @@
...
+      (buf.validate.field).repeated.items.enum.not_in = 0
+    ];
+    // The registry types to list Plugins for.
+    //
+    // If empty, all Plugins are listed regardless of registry type.
+    repeated PluginRegistryType registry_types = 4 [
+      (buf.validate.field).repeated.unique = true,
+      (buf.validate.field).repeated.items.enum.defined_only = true,
+      (buf.validate.field).repeated.items.enum.not_in = 0
+    ];
```
stefanvanburen, line 104: Thoughts on adding a `name_query` (open to suggestions on the field name) to `Filter`, similar to [`ListLabels`](https://github.com/bufbuild/registry-proto/blob/c4103f09d463d36c2cc33816b3e1e2d951a915e6/buf/registry/module/v1/label_service.proto#L139-L140)? This would allow for a client to fuzzy-filter based on user input.

mfridman: Oh, is that what we're using it for? If I input `protoc-gen-` it'll match `protoc-gen-go-*`?

stefanvanburen: yeah — I was thinking of a user entering `grp` and seeing all the plugins with the `grpc/` owner (hence maybe not using `name_query`, unless we consider "owner" as a part of the plugin's "name"), or e.g. `community/scalapb-zio-grpc`. Similar to the experience on https://buf.build/plugins now, but server-side instead of client-side.

mfridman: Gotcha, I feel like filtering plugins by organizations is covered by the `owner_refs` filter. Although you have to supply the exact organization name (or id).

I supposed we could have a `fullname_query` that is orthogonal to filtering by an organization, which could support:

- `grpc/*` (already covered by `owner_refs`)
- `grp*`
- `gatewa*`
- `json*`

I think these would be the most common types of searches. And searching by plugin name is typically what you'd want?

So maybe `name_query`, which applies to just plugin names, is the most appropriate for most cases?

buf/registry/plugin/v1beta1/commit.proto
```diff
@@ -0,0 +1,84 @@
...
+  Digest digest = 5 [(buf.validate.field).required = true];
+  // The id of the User that created this Commit on the BSR.
+  //
+  // May be empty if the User is no longer available.
+  string created_by_user_id = 6 [
+    (buf.validate.field).string.tuuid = true,
+    (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
+  ];
+  // The URL to the license of the Plugin.
+  string license_url = 7 [
```
bufdev, line 66: In https://github.com/bufbuild/bufplugin/tree/main/buf/plugin/info/v1, we have the following:

- A SPDX license ID
- A `oneof` for either the license text, or url
- Both `short` and `long` sections for documentation

Why are we differentiating in capabilities here?

buf/registry/plugin/v1beta1/commit.proto
```diff
@@ -0,0 +1,84 @@
...
+  string plugin_id = 4 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.tuuid = true
+  ];
+  // The Digest of the Commit's contents.
+  Digest digest = 5 [(buf.validate.field).required = true];
+  // The id of the User that created this Commit on the BSR.
+  //
+  // May be empty if the User is no longer available.
+  string created_by_user_id = 6 [
```
bufdev, line 61: Why or why not is `source_control_url` not a field here, when it is for the module commit?

buf/registry/plugin/v1beta1/commit.proto
```diff
@@ -0,0 +1,84 @@
...
+    (buf.validate.field).string.max_len = 255,
+    (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
+  ];
+  // The SPDX identifier of the license of the Plugin.
+  string spdx_license_id = 8;
+  // The documentation of the Plugin.
+  string doc = 10;
+}
+
+// The compression type of the content a commit provides.
```
bufdev, line 77: I don't see this referenced anywhere, what is this doing here?

buf/registry/plugin/v1beta1/digest.proto
```diff
@@ -0,0 +1,43 @@
...
+  ];
+  // The value of the Digest.
+  bytes value = 2 [(buf.validate.field).required = true];
+}
+
+// The type of Digest.
+enum DigestType {
+  DIGEST_TYPE_UNSPECIFIED = 0;
+  // The sha256 digest function.
+  DIGEST_TYPE_SHA256 = 1;
```
bufdev, line 40: Why is this here? This doesn't make sense per our conversation about digest functions.

buf/registry/plugin/v1beta1/plugin.proto
```diff
@@ -0,0 +1,142 @@
...
+    (buf.validate.field).required = true,
+    (buf.validate.field).enum.defined_only = true
+  ];
+  // The Plugin type.
+  PluginType type = 8 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).enum.defined_only = true
+  ];
+  // The configurable short description of the Plugin.
+  string description = 9 [(buf.validate.field).string.max_len = 350];
```
bufdev, line 66: Why would I want this in addition to the commit-level documentation?

buf/registry/plugin/v1beta1/plugin.proto
```diff
@@ -0,0 +1,142 @@
...
+  // PLUGIN_STATE_DEPRECATED says that the Plugin has been deprecated and should no longer be used.
+  PLUGIN_STATE_DEPRECATED = 2;
+}
+
+// The type of the Plugin.
+enum PluginType {
+  PLUGIN_TYPE_UNSPECIFIED = 0;
+  // PLUGIN_TYPE_PROTOC_CODE_GENERATION says that the Plugin is a protoc code generation plugin. It
+  // is intentionally a separate type because it does not use the pluginrpc protocol.
+  PLUGIN_TYPE_PROTOC_CODE_GENERATION = 1;
```
bufdev, line 105: I believe we explicitly said in our discussions we did not want the API representing old protoc plugins, am I misremembering? Why or why not?

buf/registry/plugin/v1beta1/plugin_service.proto
```diff
@@ -0,0 +1,217 @@
...
+  //
+  // The default value is 10.
+  uint32 page_size = 1 [(buf.validate.field).uint32.lte = 250];
+  // The page to start from.
+  //
+  // If empty, the first page is returned.
+  string page_token = 2 [(buf.validate.field).string.max_len = 4096];
+  // An optional filter to filter the list of Plugins. If unset, all Plugins for all organizations
+  // are listed, but this functionality is limited to Users with the necessary permissions.
+  Filter filter = 3;
```
bufdev, line 116: Why was this put under a nested message, when it was not in the module API? Ie what is the reasoning used here to add inconsistency to our APIs across packages?

buf/registry/plugin/v1beta1/download_service.proto
```diff
@@ -0,0 +1,70 @@
...
+  ];
+}
+
+message DownloadResponse {
+  // Content for a single version of a Plugin.
+  message Content {
+    // The Commit associated with the Content.
+    Commit commit = 1 [(buf.validate.field).required = true];
+    // Compression type of the content.
+    CompressionType compression_type = 2 [
```
bufdev, line 62: Why are you compressing data within a Protobuf message, as opposed to relying on the compression of the protocol?

buf/registry/plugin/v1beta1/code_generation_config.proto
```diff
@@ -0,0 +1,333 @@
...
+
+syntax = "proto3";
+
+package buf.registry.plugin.v1beta1;
+
+import "buf/validate/validate.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/plugin/v1beta1";
+
+// The configuration for code generation.
```
bufdev, line 23: - I don't know what this means (ie it needs documentation)
- Everything is config, why are these messages named "Config"?

buf/registry/plugin/v1beta1/plugin.proto
```diff
@@ -0,0 +1,142 @@
...
+  // PLUGIN_TYPE_CODE_GENERATION says that the Plugin is a code generation plugin.
+  PLUGIN_TYPE_CODE_GENERATION = 2;
```
bufdev, line 107: We don't have generation plugins at this time, why is this here? And why is `CodeGenerationConfig` here?

bufdev (changes requested):

buf/registry/plugin/v1beta1/plugin_service.proto
mfridman: I think having a `Filter` message aids in readability, especially as the number of items you filter by increases.

Having a single nested message to hold all the optional filters felt more appropriate instead of mixing it with other fields like page_size, page_token and order (sorting)

Maybe we could update modules to have a message `Filter` and make this a consistent pattern?

mfridman:
@bufdev did you have an opinion on @doriable 's proposal in https://github.com/bufbuild/registry-proto/pull/132#issuecomment-2407967961

Afaics the initial module API was intentionally designed to isolate all the module-related definitions. Did we want to revisit this with the addition of plugins?

buf/registry/plugin/v1beta1/digest.proto
mfridman: This hinged on the new API exposing docker-based protoc plugins. When fetching those plugins it'd be odd to exclude the digest.

If we remove docker-based protoc plugins from this API then we don't need this.

buf/registry/plugin/v1beta1/download_service.proto
mfridman: Removed all references to compression b687e130

buf/registry/plugin/v1beta1/commit.proto
bufdev: I would rather registry-proto be distinct from bufplugin

buf/registry/plugin/v1beta1/commit.proto
bufdev: What is the answer to why or why not this field was not present in this PR already? This determines how you think about this solution.

buf/registry/plugin/v1beta1/digest.proto
bufdev: I don't understand why you'd name the docker-based plugin digest type "sha256" - we have had conversations about how the hash algorithm is just one component of a digest algorithm. Names such as "docker_sha256", "docker1", or the like, would make more sense.

buf/registry/plugin/v1beta1/plugin.proto
bufdev: Perhaps we could remove both. Do you have a sense of how many descriptions are set for modules? If not many, I rather just wipe them all.

buf/registry/plugin/v1beta1/plugin.proto
bufdev: We had a conversation about this - can you remind me on what we concluded and why?

buf/registry/plugin/v1beta1/plugin_service.proto
bufdev: I think the conclusion I take here is that there isn't a strong reason to add inconsistency - readability is a judgement call. Judgement calls in style take a back seat to consistency in APIs, and the non-nested-message version has already shipped in v1, therefore please match the existing pattern.

buf/registry/plugin/v1beta1/plugin.proto
bufdev: Do not remove protoc plugins yet - the question about why or why not it is in there is not answered.

bufdev: > I thought the intention was to have a new upload service that would support both? Otherwise, we'll need to implement a stable buf command that relies on v1alpha1 CreateCuratedPlugin (docker) and v1 Upload (wasm),

Regardless of whether or not the new v1 API understands both legacy protoc and new check/generate plugins, the intention was definitely not for the new v1 API to support upload of both. Upload of legacy plugins should be left to the v1alpha1 API, or split out into a separate package. Many stable buf commands already rely on v1alpha1 APIs as it is, but regardless we should not dirty the new v1 `plugin` package with legacy plugin uploads based on docker.

bufdev:
> @bufdev did you have an opinion on @doriable 's proposal in https://github.com/bufbuild/registry-proto/pull/132#issuecomment-2407967961

I think in principle, I'd agree with it - we're trying to merge our understanding of how resources are versioned (commits, labels), and it would be nice if the API reflected that. However, when you actually look at what that means implementation-wise, it will be a significant refactor without much/any gain product-wise - we're talking weeks of work here without a payoff. You'll have to keep around all the existing `module.v1` and `module.v1beta1` types as they have shipped in v1 and are used by the CLI - removing the services isn't an option. I would instead strive to keep the common types as closely-matched as possible (see `source_control_url` et all above), giving us the ability to merge our concept of resource in a `v2` set of API packages in the future.

buf/registry/plugin/v1beta1/commit.proto
mfridman: I excluded it because it (initially) made less sense for plugins. For modules, you commit proto files, license, readme to vcs. There's a direct link between the committed files, a commit and the resulting contents pushed up to the BSR.

The GitHub action pushes each vcs commit (regardless of proto changes) to the BSR.

For plugins, you're typically not building and committing the compiled wasm module into vcs. And you're very unlikely to be publishing for every vcs commit, instead, you'd be doing a bunch of work on the plugin and when it's ready tag it and release. The model felt slightly different here.

I agree we should add this field to the `Commit` response object though. If a plugin from the BSR can point to the corresponding source code from which it was built then it's a net win.

buf/registry/plugin/v1beta1/plugin.proto
mfridman: It's a _very_ small amount. It's also not directly surfaced in the BSR outside of the repository settings page.

Isn't this a useful field that allows users to describe their module or plugin as a whole?

buf/registry/plugin/v1beta1/plugin_service.proto
mfridman: Done. d9ba013e

buf/registry/plugin/v1beta1/plugin.proto
bufdev: In a `doc.short` and `doc.long` world (which I think Modules could migrate to as well), `doc.short becomes `description`.

To enable this easily, I'd be in favor of re-defining `Doc` on `bufplugin` to make `short` and `long` distinct, ie you do *not* do `doc.short + "\n\n" + doc.long`, instead they are completely disjoint. In this world, for modules, `README.md/buf.md` maps to `doc.long`, `description` maps to `doc.short`.

BSR outside of repo settings surfaces `doc.short` of latest commit on default label.

Cc @perezd

buf/registry/plugin/v1beta1/digest.proto
mfridman: Fair, this was poorly named based on our understanding, `docker_sha256` would have been more appropriate.

My initial thought was the element of least surprise since I'd wager when most users think about "docker digests" they are thinking of the sha26 digest.

buf/registry/plugin/v1beta1/plugin.proto
mfridman: Okay great.

> Upload of legacy plugins should be left to the v1alpha1 API

@emcfarlane relevant when we get to `buf plugin push --image-{digest,config}`.

buf/registry/plugin/v1beta1/plugin.proto
mfridman: We initially said we wanted to unify all plugins into a common "plugin" entity (starting with protoc and check) since they share the same namespace and would reduce FE complexity. But then quickly realized we'd have to think through all the generated sdk config definitions and it was unclear how that would evolve when adding future generate plugins.

Furthermore, there were outstanding questions on whether the old protoc plugins would be updated to commits/labels in the UI (they are currently version/revision-based). This was after we decided that _new_ plugins will adopt a commit/label model instead of version/revision.

I do not remember us making a final decision on the API though, @emcfarlane @smallsamantha if you remember differently.

I propose we focus on check plugins in this new API to unblock new feature work like:

1. uploading check plugins and introducing `buf plugin push`
2. downloading check plugins and referencing them in `buf.yaml` files, unlocking shareable custom lint rules
3. introducing BSR features that use check plugins and custom lint rules
4. introducing the `buf registry plugin` set of commands, to have parity with `buf registry module`

buf/registry/plugin/v1beta1/commit.proto
mfridman: I added a `buf/registry/plugin/v1beta1/plugin_info.proto` file that copies the `License` and `Doc` definitions from bufplugin, which are then exposed on the response Commit

https://github.com/bufbuild/registry-proto/blob/d9ba013e876d46e98e9f2307c4b31a6070ea62a1/buf/registry/plugin/v1beta1/commit.proto#L66-L69

buf/registry/plugin/v1beta1/plugin.proto
mfridman: There's a bit of a discrepancy in how these work?

- The module readme (long) is on the commit. The description (short) is on the module entity.
- The plugin long and short are both on the commit.

Would users update the module description via the UI/API, and the plugin description by pushing a commit?

To make things work equally, we could ignore plugin info `doc.short` and only store `doc.long`. Leaving long descriptions on the commit, and short descriptions on the module/plugin entity.

buf/registry/plugin/v1beta1/upload_service.proto
```diff
@@ -0,0 +1,64 @@
...
+    // or will be updated to point to the pre-existing Commit for the Reference.
+    //
+    // If no labels are referenced, the default Label for the Plugin is used.
+    //
+    // If the Labels do not exist, they will be created.
+    // If the Labels were archived, they will be unarchived.
+    repeated ScopedLabelRef scoped_label_refs = 3;
+  }
+  // The Contents of all references.
+  repeated Content contents = 1 [(buf.validate.field).repeated.min_items = 1];
```
mfridman, line 57: Plugins might get quite large, so we should consider the batched nature of this API.

Unlike modules, users will likely be pushing a single plugin. Furthermore, I don't think we have plans for the `buf plugin push` command to support multiple plugin uploads.

Lastly, there's a cost in keeping these in memory. That was one of the motivations for compressing the content payload, because we can keep a smaller footprint in memory. E.g., the ml plugin compressed compressed 16MB -> 3MB. We also need to invoke and validate each plugin.

Options:

1. Keep this as repeated, but add a validation rule for exactly one
2. Remove repeated, allow 1 upload at a time

bufdev: Every other API in registry-proto operates on batches. In the absence of a concrete performance problem, there is no current reason to change the API to be inconsistent with other APIs. Therefore, no change here is required at this time.

buf/registry/plugin/v1beta1/plugin.proto
bufdev: I don't understand the proposal you just outlined - to be clear on mine:

- Delete `Module.description` entirely.
- Add `Doc` to both the module and plugin packages.
- Make it so that `doc.short` is not concat'ed with `doc.long` in the Bufplugin API, ie if you want the content of `doc.short` to be shown in a long description of a plugin, it must be copied to `doc.long` explicitly.
- Delete `Plugin.description` in this PR.
- Derive the Module/Plugin description in the BSR UI from the `doc.short` field of the commit on the default label.
- Derive the `README.md` from `doc.long`.

bufdev: Alternatively, we could kill `Doc` entirely, and instead have rields `description` and `documentation` added to https://github.com/bufbuild/bufplugin/blob/main/buf/plugin/info/v1/plugin_info.proto#L37 and `module.v1.Commit/plugin.v1.Commit`.

buf/registry/plugin/v1beta1/commit.proto
bufdev: It seems to me that there's no reason to differentiate here - I would add `buf plugin push --source-control-url`

buf/registry/plugin/v1beta1/plugin.proto
```diff
@@ -0,0 +1,142 @@
...
+  ];
+  // The configurable short description of the Plugin.
+  string description = 9 [(buf.validate.field).string.max_len = 350];
+  // The configurable source URL of the Plugin.
+  string source_url = 10 [
+    (buf.validate.field).string.uri = true,
+    (buf.validate.field).string.max_len = 255,
+    (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
+  ];
+  // Managed indicates whether a Plugin is managed by the BSR.
```
bufdev, line 73: This field doesn't make sense to me - what do you mean "they will be overwritten?" What do you mean "automatically updated"? These arent entity properties, this is a function of how users use the API.

This field feels incorrect and like it shouldn't be here.

buf/registry/plugin/v1beta1/plugin_service.proto
```diff
@@ -0,0 +1,194 @@
...
+    // The configurable description of the Plugin.
+    string description = 4 [(buf.validate.field).string.max_len = 350];
+    // The configurable source URL of the Plugin.
+    string source_url = 5 [
+      (buf.validate.field).string.uri = true,
+      (buf.validate.field).string.max_len = 255,
+      (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
+    ];
+    // The Plugin type.
+    PluginType type = 6 [
```
bufdev, line 134: What does it mean to create a plugin of type `PLUGIN_TYPE_PROTOC`? More generally, what does it mean to create a plugin without any commits? How is (or how is this not) congruent with Modules?

buf/registry/plugin/v1beta1/plugin_service.proto
```diff
@@ -0,0 +1,194 @@
...
+  string page_token = 2 [(buf.validate.field).string.max_len = 4096];
+  // The specific Organizations to list Plugins for.
+  //
+  // If empty, all Plugins for all Organizations are listed, but this functionality is limited to
+  // Users with the necessary permissions.
+  repeated buf.registry.owner.v1.OwnerRef owner_refs = 3;
+  // The collections to list Plugins for.
+  //
+  // If empty, all Plugins are listed regardless of collection.
+  repeated string collection_names = 4;
```
bufdev, line 93: Why are you referencing collection names when Collection is an entity?

buf/registry/plugin/v1beta1/plugin.proto
```diff
@@ -0,0 +1,137 @@
...
+  PLUGIN_STATE_ACTIVE = 1;
+  // PLUGIN_STATE_DEPRECATED says that the Plugin has been deprecated and should no longer be used.
+  PLUGIN_STATE_DEPRECATED = 2;
+}
+
+// The type of the Plugin.
+enum PluginType {
+  PLUGIN_TYPE_UNSPECIFIED = 0;
+  // PLUGIN_TYPE_CHECK says that the Plugin is a check plugin.
+  PLUGIN_TYPE_CHECK = 1;
```
bufdev, line 104: There was a comment about `PLUGIN_TYPE_PROTOC` that was deleted from the current view, because `PLUGIN_TYPE_PROTOC` was removed without us concluding the discussion around whether or not it should be included. I was confused when I couldn't find the discussion.

The answer to whether `PLUGIN_TYPE_PROTOC` should be included depends on a few areas:

- Can you map version/revision to label/commit?
- What does it mean to request a download of a protoc plugin? What error is returned otherwise?
- How do you explain in the docs that you cannot upload a plugin of type protoc?

buf/registry/plugin/v1beta1/commit.proto
bufdev: TODO: Depend on bufplugin

buf/registry/plugin/v1beta1/plugin.proto
bufdev: Talked offline:

- `source_url` and `description` to live on `Module` and `Plugin`.
- README and LICENSE will live on the `Commit` level for both `Module` and `Plugin`.

See https://github.com/bufbuild/bufplugin/pull/7

buf/registry/plugin/v1beta1/plugin.proto
mfridman: Removing this from the initial proposal. Will revisit this once we need this functionality.

For future, the intention is to have a way to differentiate between plugins that are custom (user-uploaded) or ones that are managed by the registry.

mfridman:
To recap:

- Make collection a well-defined entity with a read-only service to GetCollections and ListCollections. This is a many-many relationship, so we we'll have `collections_ids` on the Plugin.
- Remove the Plugin `managed` field. We don't need it initially. We will revisit it once we have Buf-managed check plugins.
- `description` and `source_url` belong to Plugin / Module entity, updateable via the UI or API.
	- `documentation` and `source_control_url` belong to commits, immutable.
	- bufbuild/bufplugin updated accordingly
- Okay to depend on bufplugin protos, pull them in as a dependency. Add PluginInfo to Commit.
- Punt on protoc plugins. Focus on check plugins. We don't need to solve this to ship check plugins.
- Potentially some renaming of the bufplugin to just plugin, unlikely to change to naming to "extensions".

buf.yaml
```diff
@@ -2,6 +2,7 @@ version: v2
 name: buf.build/bufbuild/registry
 deps:
   - buf.build/bufbuild/protovalidate
+  - buf.build/bufbuild/bufplugin
```
bufdev, line 5: Extreme nit: I would sort these alphabetically, to match the lock file

buf/registry/plugin/v1beta1/collection_service.proto
```diff
@@ -0,0 +1,37 @@
...
+  rpc GetCollections(GetCollectionsRequest) returns (GetCollectionsResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  // List Collections for a given Plugin.
+  rpc ListCollections(ListCollectionsRequest) returns (ListCollectionsResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+}
+
+message GetCollectionsRequest {}
```
bufdev, line 33: What's up with these?

buf/registry/plugin/v1beta1/commit.proto
```diff
@@ -0,0 +1,77 @@
...
+  //
+  // BSR users can navigate to this link to find source control information that is relevant to this Commit
+  // (e.g. commit description, PR discussion, authors, approvers, etc.).
+  string source_control_url = 7 [
+    (buf.validate.field).string.uri = true,
+    (buf.validate.field).string.max_len = 255,
+    (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
+  ];
+  // The plugin info of the Commit.
+  buf.plugin.info.v1.PluginInfo info = 8;
```
bufdev, line 76: `plugin_info`

buf/registry/plugin/v1beta1/commit.proto
```diff
@@ -0,0 +1,77 @@
...
+  // The URL of the source control commit that is associated with the Commit.
+  //
+  // BSR users can navigate to this link to find source control information that is relevant to this Commit
+  // (e.g. commit description, PR discussion, authors, approvers, etc.).
+  string source_control_url = 7 [
+    (buf.validate.field).string.uri = true,
+    (buf.validate.field).string.max_len = 255,
+    (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
+  ];
+  // The plugin info of the Commit.
```
bufdev, line 75: Make this a bit more. Something like "plugin information as stored directly within the plugin. This is part of the Bufplugin API, and can be returned from a plugin's GetPluginInfo implementation."

buf/registry/plugin/v1beta1/upload_service.proto
```diff
@@ -0,0 +1,73 @@
...
+    // BSR users can navigate to this link to find source control information that is relevant to this Commit
+    // (e.g. commit description, PR discussion, authors, approvers, etc.).
+    string source_control_url = 4 [
+      (buf.validate.field).string.uri = true,
+      (buf.validate.field).string.max_len = 255,
+      (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
+    ];
+  }
+  // The Contents of all references.
+  repeated Content contents = 1 [(buf.validate.field).repeated.min_items = 1];
```
bufdev, line 66: From our conversation: OK to reintroduce `CompressionType` and compression here

buf/registry/plugin/v1beta1/collection_service.proto
```diff
@@ -0,0 +1,103 @@
...
+  // Get Collections.
+  rpc GetCollections(GetCollectionsRequest) returns (GetCollectionsResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  // List Collections for a given Plugin.
+  rpc ListCollections(ListCollectionsRequest) returns (ListCollectionsResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  // Get the Collections for the given Plugins.
+  rpc GetCollectionsPlugins(GetCollectionsPluginsRequest) returns (GetCollectionsPluginsResponse) {
```
mfridman, line 36: @bufdev not sure how much you hate this, but as @emcfarlane and I flushed this model out it seemed better to remove `collectios_ids` from the Plugin entity and instead have a flattened list via this RPC.

Happy to bring back `collection_ids` on Plugin, but figured I'd add this for discussion. Say if you have PluginRefs and want to know which collection each of those belongs to you would:

1. Call this endpoint with plugin refs
2. Resolve the set of `collection_id` against `GetCollections`

buf/registry/plugin/v1beta1/upload_service.proto
mfridman: For brevity, do we also want to have compression type on the download service as well for content?

bufdev (approved):

bufdev:
I would probably keep in v1beta1 for a little bit unless you're supremely confident this is the final shape, but up to you

mfridman:
It was changed back to `v1beta1` in dea5e2a2.

I'd feel more confident marking it as v1 once we've flushed out code generation plugins.

PR #134, opened 2024-10-25
Allow setting spdx_license_id in module upload

mfridman:
This PR adds `spdx_license_id` field to module `UploadRequest`. The goal is to offer users more control over the resulting spdx license identifier that gets exposed once a module has been published. Such as in Generated SDKs.

The intention is to add a `--spdx-license-id` flag (or `--license` if we wanted to keep it short) to the `buf push` command.

For this to be useful, we need to expand the spdx expression logic to support (in addition to the SPDX License List):

> a user defined license reference denoted by the LicenseRef-[idString]
> 
> ... idstring = 1*(ALPHA / DIGIT / "-" / "." )

[Reference](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/)

### DIFF
--- a/buf/registry/module/v1/commit.proto
+++ b/buf/registry/module/v1/commit.proto
@@ -75,4 +75,10 @@ message Commit {
     (buf.validate.field).string.max_len = 255,
     (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
   ];
+  // The SPDX License Identifier of the Commit.
+  //
+  // May be empty if the SPDX License Identifier was not explicitly set during upload, if the
+  // license file was not present, or if it was not possible to detect the SPDX License Identifier
+  // from the license file.
+  string spdx_license_id = 8;
 }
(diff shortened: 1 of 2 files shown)


buf/registry/module/v1/commit.proto
```diff
@@ -75,4 +75,10 @@ message Commit {
     (buf.validate.field).string.max_len = 255,
     (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED
   ];
+  // The SPDX License Identifier of the Commit.
+  //
+  // May be empty if the SPDX License Identifier was not explicitly set during upload, if the
+  // license file was not present, or if it was not possible to detect the SPDX License Identifier
+  // from the license file.
+  string spdx_license_id = 8;
```
perezd, line 83: should we validate these strings w/ regex?

bufdev (commented):
I can't remember what the thinking was but I explicitly didn't want to do this for some reason when I looked into it https://github.com/bufbuild/registry-proto/issues/18

mfridman:
There's probably more of an argument to do add it now, especially since we allow this field to be explicitly set in [PluginInfo](https://buf.build/bufbuild/bufplugin/docs/main:buf.plugin.info.v1#buf.plugin.info.v1.PluginInfo).

bufdev:
Also added to `v1beta1` as these need to be kept in sync. I added the validation to `bufplugin` as well.

bufdev (approved):

PR #138, opened 2025-01-06
Add initial Policy specification

doriable:
This PR adds support for "check policies" as a feature in our API.

A new entity, `Policy` is added. Policies can only be owned by organizations,
and are in the same namespace as modules and plugins. A policy has immutable
versions (`Version`) and each version represents a `Configuration` of the `Policy`.

Uploading a new `Configuration` of a `Policy` will increment its `Version`. If the same
`Configuration` is uploaded, then the current latest `Version` is returned.

### DIFF
--- a/buf/registry/policy/v1beta1/configuration.proto
+++ b/buf/registry/policy/v1beta1/configuration.proto
@@ -1,0 +1,144 @@
+// Copyright 2023-2024 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.registry.policy.v1beta1;
+
+import "buf/validate/validate.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/policy/v1beta1";
+
+// The Configuration of a specific Policy at a specific Version. This consists of the built-in
+// rules version and lint, breaking, and plugin configurations required to run the Policy.
+message Configuration {
+  // The built-in rules version. This specifies which version of built-in rules to use for the Policy.
+  enum BuiltInVersion {
+    BUILT_IN_VERSION_UNSPECIFIED = 0;
+    BUILT_IN_VERSION_V1BETA1 = 1;
+    BUILT_IN_VERSION_V1 = 2;
+    BUILT_IN_VERSION_V2 = 3;
+  }
+  // A check config represents the shared configurations for checks (e.g. lint and breaking).
+  message CheckConfig {
+    // The list of check rules and/or categories used for the Policy.
+    repeated string use = 1 [
+      (buf.validate.field).repeated.items = {
+        string: {
+          min_len: 2
+          max_len: 250
+        }
+      },
+      (buf.validate.field).repeated.max_items = 250
+    ];
+    // The list of check rules and/or categories to exclude for the Policy.
+    repeated string except = 2 [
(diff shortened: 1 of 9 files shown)


buf/registry/policy/v1beta1/configuration.proto
```diff
@@ -0,0 +1,144 @@
...
+// The Configuration of a specific Policy at a specific Version. This consists of the built-in
+// rules version and lint, breaking, and plugin configurations required to run the Policy.
+message Configuration {
+  // The built-in rules version. This specifies which version of built-in rules to use for the Policy.
+  enum BuiltInVersion {
+    BUILT_IN_VERSION_UNSPECIFIED = 0;
+    BUILT_IN_VERSION_V1BETA1 = 1;
+    BUILT_IN_VERSION_V1 = 2;
+    BUILT_IN_VERSION_V2 = 3;
+  }
```
doriable, line 32: This allows the user to specify the lint/breaking rules version they are using, which is tied to their `buf.yaml` configuration version currently. Should we allow this or should all policies default to `v2` built-in rule sets?

buf/registry/policy/v1beta1/configuration.proto
```diff
@@ -0,0 +1,144 @@
...
+      }];
+    }
+    // The Plugin name.
+    Name name = 1 [(buf.validate.field).required = true];
+    // The ref to a specific Plugin commit. This could be a label name or a commit ID.
+    // If no reference is set, then at the time of running the Policy, the latest commit on
+    // the default label of the Plugin is used.
+    //
+    // If this value is present but empty, this should be treated as not present.
+    string ref = 2 [(buf.validate.field).string.max_len = 250];
```
doriable, line 122: Should we allow users to leave the `ref` unspecified, e.g.

```
plugins:
  - plugin: buf.build/bufbuild/timestamp-suffix
```

This would mean that each time this policy is run, it would use the latest commit on the default label for the plugin.

Discussing this with @mfridman, this is similar to specifying `runs-on: ubuntu-latest` for github actions runners -- for each run of the action, the user can introspect what `ubuntu-latest` is and they would get all updates/changes to `ubuntu-latest` without having to change their job configuration. This feels like a desirable UX.

buf/registry/policy/v1beta1/upload_service.proto
```diff
@@ -0,0 +1,58 @@
...
+
+message UploadRequest {
+  option (buf.registry.priv.extension.v1beta1.message).request_only = true;
+
+  // Content to upload.
+  message Content {
+    // The Policy of the reference.
+    PolicyRef policy_ref = 1 [(buf.validate.field).required = true];
+    // The Configuration of the Content.
+    Configuration configuration = 2 [(buf.validate.field).required = true];
```
doriable, line 43: Currently we are uploading just the `Configuration` structure. This would mean that if there was a comment change in the policy definition file but nothing else, for example, then no new `Version` is created. We would also not consider the policy configuration as "content" and would have no knowledge of the original source, just the `Configuration` shape.

buf/registry/policy/v1beta1/policy_service.proto
```diff
@@ -0,0 +1,194 @@
...
+  // Update exiting Policies.
+  //
+  // This operation is atomic. Either all Policies are updated or an error is returned.
+  rpc UpdatePolicies(UpdatePoliciesRequest) returns (UpdatePoliciesResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
+  // Delete existing Policies.
+  //
+  // This operation is atomic. Either all Policies are deleted or an error is returned.
+  rpc DeletePolicies(DeletePoliciesRequest) returns (DeletePoliciesResponse) {
```
cyinma, line 51: do we actually want policy able to be deleted?

buf/registry/policy/v1beta1/configuration.proto
doriable: Resolved this in our last discussion, we will be defaulting to the `v2` built-in rule set for policy checks. We are also uploading the entire config file instead of a proto message.

buf/registry/policy/v1beta1/configuration.proto
doriable: Resolved this in our last discussion, we will be allowing plugin name. This will not be captured here, since we are no longer putting this configuration into proto.

bufdev (approved):

emcfarlane (approved):

PR #142, opened 2025-04-22
Add Digests to Policies

emcfarlane:
This adds digest support to policies, following on from modules and plugins. The digest type for policies is O1.

### DIFF
--- a/buf/registry/plugin/v1beta1/commit_service.proto
+++ b/buf/registry/plugin/v1beta1/commit_service.proto
@@ -53,7 +53,7 @@ message GetCommitsRequest {
   // If this DigestType is not available, an error is returned.
   // Note that certain DigestTypes may be deprecated over time.
   //
-  // If not set, the latest DigestType is used, currently p1.
+  // If not set, the latest DigestType is used, currently P1.
   DigestType digest_type = 2 [(buf.validate.field).enum.defined_only = true];
 }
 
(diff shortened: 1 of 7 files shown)


buf/registry/plugin/v1beta1/commit_service.proto
```diff
@@ -53,7 +53,7 @@ message GetCommitsRequest {
   // If this DigestType is not available, an error is returned.
   // Note that certain DigestTypes may be deprecated over time.
   //
-  // If not set, the latest DigestType is used, currently p1.
+  // If not set, the latest DigestType is used, currently P1.
```
nicksnyder, line 56: Is this supposed to be O1 too?

buf/registry/plugin/v1beta1/label_service.proto
```diff
@@ -173,7 +173,7 @@ message ListLabelHistoryRequest {
   // If this DigestType is not available, an error is returned. Note that certain DigestTypes may be
   // deprecated over time.
   //
-  // If not set, the latest DigestType is used, currently B5.
+  // If not set, the latest DigestType is used, currently P1.
```
nicksnyder, line 176: ditto

buf/registry/plugin/v1beta1/commit_service.proto
emcfarlane: No this is in `plugin`, plugin digest type is still `p1`.

PR #147, opened 2025-05-15
Add FileDescriptorSet service

bufdev:
This replaces https://github.com/bufbuild/registry-proto/pulls in its entirety. Context is over there, but all comments I consider addressed. Specifically, the include/exclude decisions match what we do with https://buf.build/docs/bsr/module/descriptor

This does *not* currently include a mechanism to get Image-specific extensions. We want to that to use the new extensions that Google gave us in descriptor.proto, but we haven't had the time to move to those from the current `Image` format. We can add this down the road.


buf/registry/module/v1/file_descriptor_set_service.proto
```diff
@@ -0,0 +1,104 @@
...
+  // Only include specific types (and types that they reference, if exclude_imports is not set).
+  //
+  // A symbol is a package, message, enum, service, method, or extension.
+  //
+  // Types are referenced by their fully-qualified name. For example, if "Foo" is a message within
+  // the package "foo.v1", the fully-qualified name is "foo.v1.Foo". Fully-qualfied names should
+  // not start with a period, that is specify "foo.v1.Foo", not ".foo.v1.Foo".
+  //
+  // If no types are provided, all types are returned.
+  repeated string types = 5;
```
jhump, line 83: So we maybe want to have both `include_tyoes` and `exclude_types` , since that is available in the CLI (and easy to reuse for the implementation here)?

buf/registry/module/v1/file_descriptor_set_service.proto
```diff
@@ -0,0 +1,104 @@
...
+  //     paths: [
+  //       "message_type.name",
+  //       "message_type.nested_type.name"
+  //     ]
+  //   }
+  //
+  // The filters exclude_imports, exclude_source_code_info, exclude_source_retention_options and
+  // types continue to be respected even if this field is set. Specifically, if exclude_source_code_info
+  // is set, but this mask includes source_code_info, SourceCodeInfo will continue to be excluded.
+  google.protobuf.FieldMask field_mask = 6;
```
jhump, line 98: Are we adding this to a lot of RPCs? Seems like we could leave out, especially since it will be error prone. For example, the sample given for message names only can return top-level names, not names for nested types. The thing Alfred is working on in bufstream will be a lot more capable. Also, it's feels awkward to have to add this to every request when we could instead implement this in a cross-cutting way in the BSR by accepting the mask in a header.

buf/registry/module/v1/file_descriptor_set_service.proto
doriable, line 83: Should we have a protovalidate rule for this that fqn's do not start with `.`? (Same goes for `exclude_types` if we add it).

bufdev: Go for it

bufdev: Yep makes sense

buf/registry/module/v1/file_descriptor_set_service.proto
bufdev: Can remove for v1 of this

doriable (approved):

PR #148, opened 2025-06-13
Update to protovalidate oneofs

bufdev:
I know this is a source breaking change but is this a wire breaking change if I don't care about presence? Cc @mcy


mcy:
This is only a wire-breaking change insofar that if a writer wrote one of the oneof members with a zero value, you can no longer tell which one. If you make all of the original oneof members `optional` it is not wire-breaking in this case.

bufdev:
Every single `oneof` here was either:

- `required`, and in proto3 world therefore non-zero
- Request-only, and never persisted

So this is not breaking for this API.

PR #149, opened 2025-07-18
Upgrade protovalidate

unmultimedio:
Upgrade protovalidate ignore rules.

See protovalidate release notes: https://github.com/bufbuild/protovalidate-go/releases/tag/v0.14.0


buf/registry/module/v1beta1/module_service.proto
```diff
@@ -166,8 +166,7 @@ message UpdateModulesRequest {
     // The configurable URL in the description of the module.
     optional string url = 6 [
       (buf.validate.field).string.uri = true,
-      (buf.validate.field).string.max_len = 255,
-      (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+      (buf.validate.field).string.max_len = 255
```
unmultimedio, line 169: Usages of `IGNORE_IF_DEFAULT_VALUE` in this repo were only present in `optional` fields, which don't need any ignore value:

```
// For fields that track presence (e.g. adding the `optional` label in proto3),
// this a no-op and behavior is the same as the default `IGNORE_UNSPECIFIED`.
```
https://buf.build/bufbuild/protovalidate/docs/v0.14.1:buf.validate#buf.validate.Ignore

PR #156, opened 2025-09-16
Fix CI

stefanvanburen:
Noticed CI is failing because the latest version of Buf requires Go 1.24 and we don't pin a version. Remove the version pin.

Also removes cache-dependency-path, because the Makefile doesn't have version information that would result in the cache being busted.


stefanvanburen:
CI failure: https://github.com/bufbuild/registry-proto/pull/155

pkwarren (approved):

PR #162, opened 2026-01-09
Fix: address style consistency issues

rodaine:
## Summary

- Add `required` and `tuuid` validation to `Association.plugin_id`
- Add `tuuid` validation to `Association.collection_ids` items
- Add comments to `UserState` enum values
- Fix typo in `UserState` enum comment ("of the a User" → "of a User")
- Reserve field 3 in `UpdatePoliciesRequest.Value`

---

Generated with an experimental Claude skill. I had it capture the style of this repo first, before apply it back on itself to check for consistency.

### DIFF
--- a/buf/registry/policy/v1beta1/policy_service.proto
+++ b/buf/registry/policy/v1beta1/policy_service.proto
@@ -157,6 +157,7 @@ message UpdatePoliciesRequest {
       (buf.validate.field).enum.defined_only = true,
       (buf.validate.field).enum.not_in = 0
     ];
+    reserved 3;
     // The deprecation status of the Policy.
     optional PolicyState state = 4 [
       (buf.validate.field).enum.defined_only = true,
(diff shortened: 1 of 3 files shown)


buf/registry/policy/v1beta1/policy_service.proto
```diff
@@ -157,6 +157,7 @@ message UpdatePoliciesRequest {
       (buf.validate.field).enum.defined_only = true,
       (buf.validate.field).enum.not_in = 0
     ];
+    reserved 3;
```
bufdev, line 160: Where did this come from?

rodaine: There was no 3 field. Either one was removed historically or it was accidentally skipped. This ensures it won't be accidentally used in the future.

rodaine: Looking through the history, it appears to be accidental.

bufdev (approved):

PR #168, opened 2026-06-09
Add visibility fields to organization create/update/get protos

jonbodner-buf:
As a first step towards enabling private organizations, the protobuf definitions for CreateOrganizationsRequest.Value, UpdateOrganizationsRequest.Value, and Organization are updated to include a visibility field. A new enum OrganizationVisibility is also defined.

### DIFF
--- a/buf/registry/owner/v1/organization_service.proto
+++ b/buf/registry/owner/v1/organization_service.proto
@@ -124,6 +124,11 @@ message CreateOrganizationsRequest {
     //
     // If not set, the Organization will default to VERIFICATION_STATUS_UNVERIFIED.
     OrganizationVerificationStatus verification_status = 4 [(buf.validate.field).enum.defined_only = true];
+    // If not set, the Organization will default to ORGANIZATION_VISIBILITY_PUBLIC.
+    OrganizationVisibility visibility = 5 [
+      (buf.validate.field).enum.defined_only = true,
+      (buf.validate.field).enum.not_in = 0
+    ];
   }
   // The Organizations to create.
   repeated Value values = 1 [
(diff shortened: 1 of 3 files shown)


buf/registry/owner/v1/organization_service.proto
```diff
@@ -124,6 +124,11 @@ message CreateOrganizationsRequest {
     //
     // If not set, the Organization will default to VERIFICATION_STATUS_UNVERIFIED.
     OrganizationVerificationStatus verification_status = 4 [(buf.validate.field).enum.defined_only = true];
+    // If not set, the Organization will default to ORGANIZATION_VISIBILITY_PUBLIC.
+    OrganizationVisibility visibility = 5 [
+      (buf.validate.field).enum.defined_only = true,
+      (buf.validate.field).enum.not_in = 0
```
bufdev, line 130: I believe this is a breaking change, needs verification

jonbodner-buf: Is it breaking? If visibility is not specified, then we default to the current behavior.

buf/registry/owner/v1/organization_service.proto
```diff
@@ -158,6 +163,10 @@ message UpdateOrganizationsRequest {
       (buf.validate.field).enum.defined_only = true,
       (buf.validate.field).enum.not_in = 0
     ];
+    OrganizationVisibility visibility = 5 [
```
pkwarren:
```suggestion
    optional OrganizationVisibility visibility = 5 [
```

buf/registry/owner/v1/organization_service.proto
jonbodner-buf: now optional

bufdev (approved):

PR #170, opened 2026-06-16
Add schema service

ajeetdsouza:
The Schema service converts schemas / data between different formats.

### DIFF
--- a/buf/registry/schema/v1/schema_service.proto
+++ b/buf/registry/schema/v1/schema_service.proto
@@ -1,0 +1,106 @@
+// Copyright 2023-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.registry.schema.v1;
+
+import "buf/registry/schema/v1/format.proto";
+import "buf/registry/schema/v1/schema_source.proto";
+import "buf/validate/validate.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/schema/v1";
+
+// Operate on schema-derived representations of Protobuf types.
+//
+// These operations are distinct from retrieving raw descriptors: to get the
+// compiled FileDescriptorSet for a Module, use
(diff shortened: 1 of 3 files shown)


buf/registry/schema/v1/schema_service.proto
```diff
@@ -0,0 +1,106 @@
...
+  // The desired format of the converted message.
+  Format output_format = 5 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).enum.defined_only = true,
+    (buf.validate.field).enum.not_in = 0
+  ];
+  // If true, fields in payload that cannot be resolved against the schema are
+  // discarded. If false (the default), an unresolvable field causes the
+  // conversion to fail.
+  bool discard_unknown = 6;
```
ajeetdsouza, line 76: `discard_unknown` is only valid when the input is JSON. Ideally, we would create separate messages for every kind of format, but I don't anticipate more input / output formats, or very many more options, so I opted to avoid unnecessary nesting here.

buf/registry/schema/v1/schema_source.proto
```diff
@@ -0,0 +1,61 @@
...
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/schema/v1";
+
+// A source of a Protobuf schema to operate on.
+//
+// A SchemaSource is either a reference to schema already stored in the BSR, or a
+// self-contained schema provided inline by the caller.
+//
+// SchemaSources can only be used in requests, and only for read-only RPCs.
+message SchemaSource {
```
ajeetdsouza, line 32: Do we need both `resource_ref` and `file_descriptor_set`?

buf/registry/schema/v1/schema_service.proto
```diff
@@ -0,0 +1,108 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.registry.schema.v1;
```
bufdev, line 17: This shouldn't be the schema package  or SchemaService - this isn't a desriptive name of what this is

buf/registry/schema/v1/schema_service.proto
```diff
@@ -0,0 +1,108 @@
...
+// These operations are distinct from retrieving raw descriptors: to get the
+// compiled FileDescriptorSet for a Module, use
+// buf.registry.module.v1.FileDescriptorSetService.
+service SchemaService {
+  // Convert a serialized message between formats using a schema.
+  rpc Convert(ConvertRequest) returns (ConvertResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  // Get an approximate JSON Schema for a message type within a schema.
+  rpc GetJSONSchema(GetJSONSchemaRequest) returns (GetJSONSchemaResponse) {
```
bufdev, line 36: I don't directly see how these two RPCs are related

PR #171, opened 2026-06-16
Add audit service

ajeetdsouza:
AuditService adds APIs to audit events. Events are actions, performed by an actor on a resource.

### DIFF
--- a/buf/registry/audit/v1/action.proto
+++ b/buf/registry/audit/v1/action.proto
@@ -1,0 +1,207 @@
+// Copyright 2023-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.registry.audit.v1;
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/audit/v1";
+
+// Action is the set of actions that the BSR records in its audit log.
+enum Action {
+  ACTION_UNSPECIFIED = 0;
+  // An Organization was created.
+  ACTION_ORGANIZATION_CREATED = 1;
+  // An Organization was deleted.
+  ACTION_ORGANIZATION_DELETED = 2;
+  // A User was added as a member of an Organization.
+  ACTION_ORGANIZATION_MEMBER_ADDED = 3;
+  // The role of a member of an Organization was changed.
+  //
+  // Superseded by ACTION_ORGANIZATION_MEMBER_ROLES_CHANGED. Only present on historical
+  // Events.
+  ACTION_ORGANIZATION_MEMBER_ROLE_CHANGED = 4 [deprecated = true];
+  // The roles of a member of an Organization were changed.
+  ACTION_ORGANIZATION_MEMBER_ROLES_CHANGED = 50;
+  // A User was removed as a member of an Organization.
+  ACTION_ORGANIZATION_MEMBER_REMOVED = 5;
+  // An identity-provider group mapping was added to an Organization.
+  ACTION_ORGANIZATION_IDP_GROUP_ADDED = 21;
+  // An identity-provider group mapping of an Organization was updated.
+  ACTION_ORGANIZATION_IDP_GROUP_UPDATED = 59;
+  // An identity-provider group mapping was removed from an Organization.
+  ACTION_ORGANIZATION_IDP_GROUP_REMOVED = 22;
+  // The resource visibility control setting of an Organization was changed.
+  ACTION_ORGANIZATION_RESOURCE_VISIBILITY_CONTROL_CHANGED = 81;
+  // The uniqueness policy of an Organization was enabled.
+  ACTION_ORGANIZATION_UNIQUENESS_POLICY_ENABLED = 82;
+  // The uniqueness policy of an Organization was disabled.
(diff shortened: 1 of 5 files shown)


buf/registry/audit/v1/action.proto
```diff
@@ -0,0 +1,207 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.registry.audit.v1;
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/audit/v1";
+
+// Action is the set of actions that the BSR records in its audit log.
+enum Action {
```
ajeetdsouza, line 22: Copied from: https://github.com/bufbuild/buf/blob/32e150f9413f0d56cde7735d8b9b9fa6ef1867c8/proto/buf/alpha/audit/v1alpha1/event.proto#L84

This renames `EVENT_TYPE_*` to `ACTION_*`, and renames `REPOSITORY` to `MODULE`. This will create some inconsistency with the value we actually store in the database, but since we no longer use the term "repository", having this in a public API is likely to be confusing.

buf/registry/audit/v1/resource.proto
```diff
@@ -0,0 +1,109 @@
...
+  // Organization members, and so on) and enables owner-scoped audit queries. It is empty for
+  // instance-level resources such as the server.
+  string owner_id = 4 [
+    (buf.validate.field).string.tuuid = true,
+    (buf.validate.field).ignore = IGNORE_IF_ZERO_VALUE
+  ];
+}
+
+// ResourceType is the type of resource affected by an audited action.
+enum ResourceType {
```
ajeetdsouza, line 58: Copied from: https://github.com/bufbuild/buf/blob/32e150f9413f0d56cde7735d8b9b9fa6ef1867c8/proto/buf/alpha/audit/v1alpha1/event.proto#L44

buf/registry/audit/v1/event.proto
```diff
@@ -0,0 +1,75 @@
...
+  Resource resource = 5 [(buf.validate.field).required = true];
+  // Metadata about the request that triggered the action.
+  RequestMetadata request_metadata = 6;
+  // Action-specific structured details.
+  //
+  // The shape of details is keyed by action; for example, ACTION_ORGANIZATION_MEMBER_ADDED
+  // carries the granted role and its source. details is intended for human and agent
+  // consumption and for forward compatibility, allowing new fields to be carried without a
+  // schema change.
+  google.protobuf.Struct details = 7;
```
ajeetdsouza, line 74: The Event payload in the internal API is a oneof with a huge list of types: https://github.com/bufbuild/buf/blob/9c4f1ea068033b314a2a45af405dc262e4ba335f/proto/buf/alpha/audit/v1alpha1/event.proto#L202 - I've used a Struct for now, but I think we should make the whole list of types part of the public API. A second opinion on this would be great

oliversun9: I think we should have a list of oneofs similar to v1alpha1, which is typed.

buf/registry/audit/v1/event.proto
```diff
@@ -0,0 +1,75 @@
...
+// Event is an immutable record of an action that occurred within the BSR.
+//
+// Every Event answers four questions: when the action occurred (create_time), who performed it
+// (actor), what was done (action), and which resource was affected (resource). Action-specific
+// details are carried in the details field.
+message Event {
+  option (buf.registry.priv.extension.v1beta1.message).response_only = true;
+
+  // RequestMetadata captures context about the request that triggered the action.
+  message RequestMetadata {
```
oliversun9, line 38: `EventMetadata` in `v1alpha1` has an ip field. Do we need it here?

buf/registry/audit/v1/audit_service.proto
```diff
@@ -0,0 +1,127 @@
...
+  // If not set, there is no upper bound.
+  google.protobuf.Timestamp end_time = 4;
+  // Filter to events performed by any of the given Users.
+  //
+  // If empty, events from all actors are returned.
+  repeated buf.registry.owner.v1.UserRef actor_refs = 5 [(buf.validate.field).repeated.max_items = 250];
+  // Filter to events affecting any of the resources with these ids.
+  //
+  // If empty, events for resources with any id are returned.
+  repeated string resource_ids = 6 [
```
oliversun9, line 92: I don't think `resource_ids` can be used without a resource type. This field could be `resouce_refs` where each ref is type + id.

buf/registry/audit/v1/audit_service.proto
```diff
@@ -0,0 +1,127 @@
...
+    (buf.validate.field).repeated.max_items = 250,
+    (buf.validate.field).repeated.items.string.tuuid = true
+  ];
+  // Filter to events affecting resources owned by any of these Users or Organizations.
+  //
+  // If empty, events for resources with any owner are returned.
+  repeated buf.registry.owner.v1.OwnerRef resource_owner_refs = 7 [(buf.validate.field).repeated.max_items = 250];
```
oliversun9, line 99: An `Event` doesn't have a resource owner id field so I'm not sure if we need this field. There is also `RESOURCE_TYPE_SERVER`, which doesn't fit in really.

buf/registry/audit/v1/audit_service.proto
oliversun9: Perhaps
```
message ResourceFilter {
  ResourceType resource_type = 1; // required
  string resource_id = 2; // optional, if empty, all resources of that type are selected.
}
```

buf/registry/audit/v1/event.proto
ajeetdsouza: I don't think we store or return IPs any more.

buf/registry/audit/v1/audit_service.proto
ajeetdsouza: Agreed, I removed the filter for now. We can add it back later if we need it.

buf/registry/audit/v1/audit_service.proto
ajeetdsouza: Removed it.

ajeetdsouza:
I've merged a version of this change directly into `iam-v1`.

PR #176, opened 2026-06-16
Add generation service

ajeetdsouza:


ajeetdsouza:
We'll tackle codegen later.

PR #181, opened 2026-06-22
Use ModuleRef and OrganizationRef when referring to a module or org in the recommended SDK Request API calls

jonbodner-buf:


buf/registry/admin/v1/instance_service.proto
```diff
@@ -122,7 +122,7 @@ message AddRecommendedSDKRequest {
   // The Plugin that generates the SDK to recommend.
   buf.registry.plugin.v1beta1.PluginRef plugin_ref = 1 [(buf.validate.field).required = true];
   // The level at which to recommend the SDK.
-  SDKRecommendationLevel recommendation_level = 2 [(buf.validate.field).required = true];
+  SDKRecommendationLevelRequest recommendation_level = 2 [(buf.validate.field).required = true];
```
bufdev: `*Request` is only used for RPC request and response objects, and is never nested

buf/registry/admin/v1/instance_service.proto
```diff
@@ -131,7 +131,7 @@ message RemoveRecommendedSDKRequest {
   // The Plugin that generates the SDK to stop recommending.
   buf.registry.plugin.v1beta1.PluginRef plugin_ref = 1 [(buf.validate.field).required = true];
   // The level at which to stop recommending the SDK.
-  SDKRecommendationLevel recommendation_level = 2 [(buf.validate.field).required = true];
+  SDKRecommendationLevelRequest recommendation_level = 2 [(buf.validate.field).required = true];
```
bufdev: Requests may never be part of responses, and this message also contains `*Ref` messages, which are request only

jonbodner-buf: That is a Request, not a response. Full context:

```
message RemoveRecommendedSDKRequest {
  // The Plugin that generates the SDK to stop recommending.
  buf.registry.plugin.v1beta1.PluginRef plugin_ref = 1 [(buf.validate.field).required = true];
  // The level at which to stop recommending the SDK.
  SDKRecommendationLevelRequest recommendation_level = 2 [(buf.validate.field).required = true];
}
```

The only usage of `RemoveRecommendedSDKRequest` is:

```
  // Remove an SDK from the list of recommended SDKs at a given level.
  rpc RemoveRecommendedSDK(RemoveRecommendedSDKRequest) returns (RemoveRecommendedSDKResponse) {
  }
```

buf/registry/admin/v1/instance_service.proto
jonbodner-buf: Is there another suggested name to use when there's one `SDKRecommendationLevel` for the response and another for the request? I had used `SDKRecommendationLevel` for both the request and the response, but was told to use ModuleRef and OrganizationRef for requests, necessitating a new message type.

jonbodner-buf: Or do I just flatten it into the parent message?

buf/registry/admin/v1/recommended_sdk.proto
```diff
@@ -82,3 +84,25 @@ message SDKRecommendationLevel {
     Module module = 3;
   }
 }
+
+// The level at which an SDK is recommended.
+//
+// An SDK can be recommended at more than one level at once; for example, an SDK
+// recommended at the instance level may also be recommended for a specific Module.
+message SDKRecommendationLevelRequest {
```
oliversun9: `SDKRecommendationLevelRef` perhaps?

buf/registry/admin/v1/recommended_sdk.proto
```diff
@@ -82,3 +84,24 @@ message SDKRecommendationLevel {
...
+message SDKRecommendationLevelRef {
+  // The SDK is recommended across the entire instance.
+  //
+  // Eager generation is enqueued for the SDK across all Modules on the instance.
+  message Instance {}
+
+  oneof level {
+    option (buf.validate.oneof).required = true;
+    // The SDK is recommended across the entire instance.
+    Instance instance = 1;
```
nicksnyder, line 101: Why is this called `Instance` here and `Server` in `SDKRecommendationLevel`? Seems like names should be consistent.

jonbodner-buf: I'm pretty sure it's called `Instance` in both:

```
// The level at which an SDK is recommended.
//
// An SDK can be recommended at more than one level at once; for example, an SDK
// recommended at the instance level may also be recommended for a specific Module.
message SDKRecommendationLevel {
  // The SDK is recommended across the entire instance.
  //
  // Eager generation is enqueued for the SDK across all Modules on the instance.
  message Instance {}
  // The SDK is recommended across a single Organization.
  //
  // Eager generation is enqueued for the SDK across all Modules owned by the Organization.
  message Organization {
    // The id of the Organization the SDK is recommended for.
    string organization_id = 1 [
      (buf.validate.field).string.tuuid = true
    ];
  }
  // The SDK is recommended for a single Module.
  //
  // Eager generation is enqueued for the SDK within the Module.
  message Module {
    // The id of the Module the SDK is recommended for.
    string module_id = 1 [
      (buf.validate.field).string.tuuid = true
    ];
  }
  oneof level {
    option (buf.validate.oneof).required = true;
    // The SDK is recommended across the entire instance.
    Instance instance = 1;
    // The SDK is recommended across a single Organization.
    Organization organization = 2;
    // The SDK is recommended for a single Module.
    Module module = 3;
  }
}

// The level at which an SDK is recommended.
//
// An SDK can be recommended at more than one level at once; for example, an SDK
// recommended at the instance level may also be recommended for a specific Module.
message SDKRecommendationLevelRef {
  // The SDK is recommended across the entire instance.
  //
  // Eager generation is enqueued for the SDK across all Modules on the instance.
  message Instance {}

  oneof level {
    option (buf.validate.oneof).required = true;
    // The SDK is recommended across the entire instance.
    Instance instance = 1;
    // The SDK is recommended across a single Organization.
    buf.registry.owner.v1.OrganizationRef organization_ref = 2;
    // The SDK is recommended for a single Module.
    buf.registry.module.v1.ModuleRef module_ref = 3;
  }
}
```

nicksnyder: Sorry, I was looking at the version in core, nevermind.

buf/registry/admin/v1/instance_service.proto
nicksnyder: I think this concern has been addressed with the latest diff.

buf/registry/admin/v1/instance_service.proto
nicksnyder: I think this concern has been addressed with the latest diff.

nicksnyder (approved):

PR #184, opened 2026-06-26
Move recommended SDK from admin to plugin

jonbodner-buf:


bufdev:
This seemed right on admin, what motivated this?

PR #185, opened 2026-06-26
Rename admin to settings

jonbodner-buf:


buf/registry/audit/v1/event.proto
```diff
@@ -16,13 +16,13 @@ syntax = "proto3";
...
 package buf.registry.audit.v1;
 
-import "buf/registry/admin/v1/instance_settings.proto";
 import "buf/registry/iam/v1/module_contributor.proto";
 import "buf/registry/iam/v1/organization_membership.proto";
 import "buf/registry/module/v1/module.proto";
 import "buf/registry/plugin/v1/plugin.proto";
 import "buf/registry/policy/v1/policy.proto";
 import "buf/registry/resource/v1/resource_visibility_control.proto";
+import "buf/registry/settings/v1/instance_settings.proto";
```
bufdev, line 25: Still named instance

jonbodner-buf: Some settings are instance level. The messages for those are defined in that file. Is there another name I should use for settings that apply to an instance?

PR #188, opened 2026-07-07
Add labels to webhook events

jonbodner-buf:
Once this is merged, https://github.com/bufbuild/core/pull/22998 needs to be updated to point at iam-v1.

### DIFF
--- a/buf/registry/webhook/v1/event.proto
+++ b/buf/registry/webhook/v1/event.proto
@@ -62,4 +62,6 @@ message ModuleCommitPushedPayload {
   buf.registry.module.v1.Module module = 1 [(buf.validate.field).required = true];
   // The Commit that was pushed.
   buf.registry.module.v1.Commit commit = 2 [(buf.validate.field).required = true];
+  // The labels associated with the commits
+  repeated string label_names = 3;
 }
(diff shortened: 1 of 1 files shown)


buf/registry/webhook/v1/event.proto
```diff
@@ -62,4 +62,6 @@ message ModuleCommitPushedPayload {
   buf.registry.module.v1.Module module = 1 [(buf.validate.field).required = true];
   // The Commit that was pushed.
   buf.registry.module.v1.Commit commit = 2 [(buf.validate.field).required = true];
+  // The labels associated with the commits
```
bufdev: Matching rest of file: full sentences, and Commit is capitalized.

There may be validation done as well for labels, I'm not sure.

jonbodner-buf: I considered a min length of 1, but nothing else came to mind.

bufdev: https://github.com/bufbuild/registry-proto/blob/main/buf/registry/module/v1/label.proto#L49 documents the current validation for label names

jonbodner-buf: From the other message types, it looks like the only applicable rule is max_len = 250. I'll add it.

Is https://github.com/bufbuild/registry-proto/blob/b8407d94d9f1b1d003ce187a1aa17a2c6446611d/buf/registry/module/v1/label.proto#L170 missing a validation?

bufdev (approved):

PR #195, opened 2026-08-18
Update ScopedPermission.permissions to be repeated

pkwarren:
When creating a scoped token, it is typical to list multiple permissions for a single scope (resource). Update `permission` to `permissions` and change the field to repeated to better match the shape of requests to create tokens.

### DIFF
--- a/buf/registry/iam/v1/token.proto
+++ b/buf/registry/iam/v1/token.proto
@@ -77,23 +77,25 @@ message Token {
   repeated ScopedPermission scoped_permissions = 7 [
     (buf.validate.field).repeated.max_items = 250,
     (buf.validate.field).cel = {
-      id: "scoped_permissions.unique"
-      message: "scoped permissions must have unique scope and permission pairs"
-      expression: "this.map(sp, (has(sp.scope.instance) ? 'instance' : string(sp.scope.resource.resource_type) + '/' + sp.scope.resource.resource_id) + '|' + sp.permission).unique()"
+      id: "scoped_permissions.unique_scope"
+      message: "scoped permissions must have unique scopes"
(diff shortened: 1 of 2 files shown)


buf/registry/iam/v1/token.proto
```diff
@@ -77,23 +77,25 @@ message Token {
...
+      expression: "this.map(sp, has(sp.scope.instance) ? 'instance' : string(sp.scope.resource.resource_type) + '/' + sp.scope.resource.resource_id).unique()"
     }
   ];
 }
 
-// A permission that applies within a specific Scope.
+// A set of permissions that apply within a specific Scope.
 //
 // ScopedPermissions restrict a Token to a subset of its owning User's permissions.
 message ScopedPermission {
```
oliversun9: should this be `ScopedPermissions`?

bufdev:
This is breaking but I assume not deployed? Non-breaking would be to reserve 2 and permission

buf/registry/iam/v1/token.proto
jonbodner-buf: Or maybe `ScopePermissions`, since it's a single `scope` paired with 1 - 100 `permissions`.

pkwarren:
> This is breaking but I assume not deployed? Non-breaking would be to reserve 2 and permission

Yes not in use yet.

bufdev (approved):

buf/registry/iam/v1/token.proto
pkwarren: Updated in 69b7551. Had to rename the wrapper in `UpdateTokensRequest` to match.

oliversun9 (approved):

PR #197, opened 2026-08-20
Add protos for a v1 trust credentials API

DMarby:
This adds a service for trust credentials, which allows setting up a trust relationship to an external provider, so that it can mint short-lived tokens for a bot user.

This allows eg a Github Actions workflow to access the BSR without the user having to create a long-lived API token, by exchanging a signed JWT by the configured issuer for a short-lived BSR token.

https://docs.github.com/en/actions/concepts/security/openid-connect#how-oidc-integrates-with-github-actions

### DIFF
--- a/buf/registry/iam/v1/trust_credential.proto
+++ b/buf/registry/iam/v1/trust_credential.proto
@@ -1,0 +1,206 @@
+// Copyright 2023-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.registry.iam.v1;
+
+import "buf/registry/iam/v1/token.proto";
+import "buf/registry/owner/v1/user.proto";
+import "buf/registry/priv/extension/v1beta1/extension.proto";
+import "buf/validate/validate.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/iam/v1";
+
+// A TrustCredential on the BSR.
+//
+// A TrustCredential is a trust rule that allows a workload to authenticate as its owning User by
+// presenting an OIDC token signed by a WorkloadIdentityIssuer, instead of a Token secret.
+//
+// A workload's OIDC token is accepted only if its sub claim matches subject and every CustomClaim
+// matches, so a TrustCredential reaches exactly as far as those allow. The subject is what pins the
+// credential to a workload; CustomClaims narrow it further.
+//
+// A TrustCredential's name, issuer_url, subject, CustomClaims, and ScopedPermissions may
+// be updated after creation. All other properties are immutable.
+//
+// Exchanging a workload's OIDC token mints a Token that authenticates as the owning User and that
+// carries the TrustCredential's ScopedPermissions. Deleting a TrustCredential revokes the
+// Tokens already minted through it. Updating one does not: a Token keeps the ScopedPermissions it
+// was minted with until it expires, so a narrowed TrustCredential governs only the Tokens minted
+// after the change.
+message TrustCredential {
+  option (buf.registry.priv.extension.v1beta1.message).response_only = true;
+
+  // The id of the TrustCredential.
+  string id = 1 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.tuuid = true
+  ];
+  // The time the TrustCredential was created.
+  google.protobuf.Timestamp create_time = 2 [(buf.validate.field).required = true];
+  // The last time the TrustCredential was updated.
+  google.protobuf.Timestamp update_time = 3 [(buf.validate.field).required = true];
+  // The name of the TrustCredential.
+  //
+  // Names are unique among the TrustCredentials owned by a User.
+  string name = 4 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.min_len = 3,
+    (buf.validate.field).string.max_len = 256
+  ];
+  // The id of the User that owns the TrustCredential.
+  //
+  // Tokens minted through the TrustCredential authenticate as this User.
+  string user_id = 5 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.tuuid = true
+  ];
+  // The WorkloadIdentityIssuer whose OIDC tokens the TrustCredential accepts.
+  //
+  // WORKLOAD_IDENTITY_ISSUER_CUSTOM means the issuer is not one this instance knows, and issuer_url
+  // is the operator's own.
+  WorkloadIdentityIssuer issuer = 6 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).enum.defined_only = true
+  ];
+  // The url of the WorkloadIdentityIssuer whose OIDC tokens the TrustCredential accepts.
+  //
+  // This is compared against the iss claim of the workload's OIDC token exactly. For a known
+  // issuer it is the url that issuer signs with, which the instance fills in.
+  string issuer_url = 7 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.uri = true,
+    (buf.validate.field).string.max_len = 255
+  ];
+  // The subject that a workload's OIDC token must carry in its sub claim.
+  //
+  // Its shape is the issuer's: GitHub Actions, for instance, writes
+  // "repo:acme/protos:ref:refs/heads/main", so "repo:acme/protos:*" trusts every workflow in that
+  // one repository.
+  //
+  // A "*" matches any run of characters, anywhere in the subject, and every other character must
+  // match exactly. Only a subject of nothing but wildcards is refused, since it would trust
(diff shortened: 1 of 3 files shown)


buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/iam/v1";
+
+// A TrustCredential on the BSR.
+//
+// A TrustCredential is a trust rule that allows a workload to authenticate as its owning User by
+// presenting an OIDC token signed by a WorkloadIdentityIssuer, instead of a Token secret.
+//
+// A workload's OIDC token is accepted only if its sub claim matches subject and every CustomClaim
+// matches, so a TrustCredential reaches exactly as far as those allow. The subject is what pins the
```
pkwarren:
```suggestion
// matches. The subject is what pins the
```

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+// credential to a workload; CustomClaims narrow it further.
```
pkwarren:
```suggestion
// credential to a workload and CustomClaims narrow it further.
```

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+//
+// A TrustCredential's name, issuer_url, subject, CustomClaims, and ScopedPermissions may
```
pkwarren: We should consistently use field names here (`CustomClaims` -> `custom_claims`).

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+// be updated after creation. All other properties are immutable.
+//
+// Exchanging a workload's OIDC token mints a Token that authenticates as the owning User and that
+// carries the TrustCredential's ScopedPermissions. Deleting a TrustCredential revokes the
+// Tokens already minted through it. Updating one does not: a Token keeps the ScopedPermissions it
+// was minted with until it expires, so a narrowed TrustCredential governs only the Tokens minted
+// after the change.
```
pkwarren: It might be good to mention the lifetime of these tokens is short so the impact is negligible. The reason they aren't revoked immediately is that any change to narrow the permissions would cause every in progress job using the token to fail mid-way.

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+    (buf.validate.field).required = true,
+    (buf.validate.field).enum.defined_only = true
+  ];
+  // The url of the WorkloadIdentityIssuer whose OIDC tokens the TrustCredential accepts.
+  //
+  // This is compared against the iss claim of the workload's OIDC token exactly. For a known
+  // issuer it is the url that issuer signs with, which the instance fills in.
+  string issuer_url = 7 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.uri = true,
```
pkwarren, line 86: Do we want to restrict these to a https scheme?

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+    (buf.validate.field).string.max_len = 255
+  ];
+  // The subject that a workload's OIDC token must carry in its sub claim.
+  //
+  // Its shape is the issuer's: GitHub Actions, for instance, writes
+  // "repo:acme/protos:ref:refs/heads/main", so "repo:acme/protos:*" trusts every workflow in that
+  // one repository.
+  //
+  // A "*" matches any run of characters, anywhere in the subject, and every other character must
+  // match exactly. Only a subject of nothing but wildcards is refused, since it would trust
```
pkwarren, line 98: > Only a subject of nothing but wildcards

Do we want to validate this with a protovalidate rule containing only `*`?

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+  // everything the issuer signs. There is no restriction on where a wildcard may fall: the shape of
+  // a sub claim is the issuer's to define, and a subject that is a bare numeric account id has no
```
pkwarren: > , and a subject that is a bare numeric account id has no segment to widen past.

I think we could remove this and not lose much in the comment.

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+  // is escaped by surrounding it in double quotes, so "\"open.id\".username" names the "username"
+  // field of the "open.id" object.
+  string key = 1 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.max_len = 250
+  ];
+  // The value that the claim must have.
+  //
+  // A "*" matches any run of characters and every other character must match exactly. A value of
+  // only wildcards is refused, as it would place no restriction on the claim at all.
```
pkwarren, line 160: Again would be a good protovalidate rule to prevent all wildcards

Do we want to support a `CustomClaim` that simply verifies that the claim is found in the token (and not care about the contents)? If so a wildcard by itself could be useful.

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,206 @@
...
+  string value = 2 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.max_len = 1000
+  ];
+}
+
+// A WorkloadIdentityIssuer whose OIDC tokens an instance accepts in exchange for a Token.
+//
+// A known issuer names a platform this instance already understands, so a TrustCredential naming
+// one does not carry a url: the instance supplies the url that platform signs with.
```
pkwarren, line 175: This feels like a useful protovalidate rule - only allow an `issuer_url` if the issuer is custom.

buf/registry/iam/v1/trust_credential_service.proto
```diff
@@ -0,0 +1,308 @@
...
+option go_package = "buf.build/gen/go/bufbuild/registry/protocolbuffers/go/buf/registry/iam/v1";
+
+// Operate on TrustCredentials.
+service TrustCredentialService {
+  // Get TrustCredentials by id or name.
+  rpc GetTrustCredentials(GetTrustCredentialsRequest) returns (GetTrustCredentialsResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+    option (buf.registry.priv.extension.v1beta1.method).required_permission = "bsr.trust-credential.get";
+  }
+  // List TrustCredentials, usually for a User.
```
pkwarren: Not sure if "usually for a User" is valid here as `user_refs` has min items of 1.

buf/registry/iam/v1/trust_credential_service.proto
```diff
@@ -0,0 +1,308 @@
...
+    //
+    // For a known issuer the instance supplies the url that platform signs with, and issuer_url
+    // must be empty. For WORKLOAD_IDENTITY_ISSUER_CUSTOM, issuer_url is required.
+    WorkloadIdentityIssuer issuer = 3 [
+      (buf.validate.field).required = true,
+      (buf.validate.field).enum.defined_only = true
+    ];
+    // The url of the OIDC issuer whose tokens the TrustCredential accepts.
+    //
+    // Required when issuer is WORKLOAD_IDENTITY_ISSUER_CUSTOM, and must be empty otherwise. It must
```
pkwarren, line 152: This constraint should be modeled as a message level protovalidate rule.

buf/registry/iam/v1/trust_credential_service.proto
```diff
@@ -0,0 +1,308 @@
...
+    // be https and carry no credentials, query, or fragment. It is compared against the iss claim
```
pkwarren, line 153: Requiring `https` should be a field level constraint.

buf/registry/iam/v1/trust_credential_service.proto
```diff
@@ -0,0 +1,308 @@
...
+    // If not set, the name is not updated.
+    optional string name = 2 [
+      (buf.validate.field).string.min_len = 3,
+      (buf.validate.field).string.max_len = 256
+    ];
+    // The WorkloadIdentityIssuer whose OIDC tokens the TrustCredential accepts.
+    //
+    // If not set, the issuer is not updated. The same pairing as on creation applies to whatever the
+    // TrustCredential holds after the update: a known issuer carries no issuer_url, and
+    // WORKLOAD_IDENTITY_ISSUER_CUSTOM requires one.
```
pkwarren, line 283: Should model this constraint as message level.

pkwarren (commented):
Overall looks great. Mainly some feedback around proto comments and a few places where we could add some more constraints to catch validation errors.

buf/registry/iam/v1/trust_credential.proto
DMarby: I think we'll want to check values on all claims, so sounds good to add a rule 👍

pkwarren (approved):

buf/registry/iam/v1/trust_credential.proto
```diff
@@ -0,0 +1,219 @@
...
+enum WorkloadIdentityIssuer {
+  WORKLOAD_IDENTITY_ISSUER_UNSPECIFIED = 0;
+  // An OIDC issuer this instance does not know, named by its own url.
+  //
+  // The url must be https and carry no credentials, query, or fragment. It is compared against the
+  // iss claim exactly, so a trailing slash is kept if the issuer has one.
+  //
+  // This is 1 so that the known issuers below occupy a contiguous range from 2 upward and adding
+  // one is just the next number.
```
nicksnyder: Is this comment material/necessary?

nicksnyder (approved):

buf/registry/iam/v1/trust_credential.proto
DMarby: Probably not! Deleted.

PR #198, opened 2026-09-08
Update audit v1 API

pkwarren:
Add constraints to ListAuditEvents RPC. Stop reserving fields in v1 messages that don't need to preserve binary serialization. Add `go_package` to the audit API (these were the only files that lacked it).


buf/registry/audit/v1/event.proto
```diff
@@ -211,99 +225,95 @@ message Event {
...
-
   // Payload of the audited event. It specifies additional context on the event.
   oneof payload {
     PayloadOrganizationCreated organization_created = 7;
     PayloadOrganizationDeleted organization_deleted = 8;
     PayloadOrganizationMemberAdded organization_member_added = 9;
-    PayloadOrganizationMemberRolesChanged organization_member_roles_changed = 56;
+    PayloadOrganizationMemberRolesChanged organization_member_roles_changed = 10;
     PayloadOrganizationMemberRemoved organization_member_removed = 11;
-    PayloadOrganizationResourceVisibilityControlChanged organization_resource_visibility_control_changed = 87;
```
ajeetdsouza, line 225: Is there any reason we want to change the field numbers? They were written to match the field numbers in this file, for ease of comparison:

https://github.com/bufbuild/buf/blob/37122b25db97a887af7be4f5b3d19bcef3811d40/proto/buf/alpha/audit/v1alpha1/event.proto#L206

pkwarren: I can revert but since this is the v1 API it made sense to not carry forward any reserved fields. If we aren't relying on the v1alpha1 event serializing as proto to a v1 event, this seemed cleaner.

Also going forward audit events should only be created/updated here and not in v1alpha1, so they'll eventually fall out of sync.

nicksnyder: I think resetting the field numbers as part of v1 makes sense

nicksnyder (approved):

PR #199, opened 2026-09-15
Add last used time to tokens

oliversun9:
Add `last_used_time` to tokens.


buf/registry/iam/v1/token.proto
```diff
@@ -82,6 +82,13 @@ message Token {
       expression: "this.map(sp, has(sp.scope.instance) ? 'instance' : string(sp.scope.resource.resource_type) + '/' + sp.scope.resource.resource_id).unique()"
     }
   ];
+  // The last time the Token was used to authenticate a request.
```
oliversun9, line 85: We might want to allow the BSR to return an approximate of this value, i.e. server implementations may want to update this value at most once every `t` (1 minute, 10 minutes, 1 hour?) per token.

oliversun9: Not sure if we want to include this in the comments

bufdev (approved):

buf/registry/iam/v1/token.proto
```diff
@@ -82,6 +82,13 @@ message Token {
...
+  //
+  // If not present, it means the last used time is unknown. This is true for tokens
+  // created or last used before BSR started tracking the last used time.
+  //
+  // If present with zero value, it means the token has never been used.
+  optional google.protobuf.Timestamp last_used_time = 8;
```
jonbodner-buf, line 91: Do we need a validation? Can't be after the current time?

jonbodner-buf (commented):
One question

buf/registry/iam/v1/token.proto
oliversun9: If the token was used just before `GetTokens/ListTokens`, the validation might fail due to clock skew I think.

jonbodner-buf (approved):

pkwarren (approved):

PR #200, opened 2026-09-18
Add trust credential audit events

DMarby:
Adds new v1 only audit events for the trust credential service


buf/registry/audit/v1/event.proto
```diff
@@ -1162,3 +1171,56 @@ message PayloadRecommendedSdkRemoved {
   // plugin_name is the name of the plugin of the recommended SDK.
   string plugin_name = 8;
 }
+
+message PayloadTrustCredentialCreated {
+  // owner_id is the id of the owner of the trust credential.
+  string owner_id = 1 [(buf.validate.field).string.tuuid = true];
```
pkwarren: nit: Should this be `user_id` to match the trust credential message?

buf/registry/audit/v1/event.proto
```diff
@@ -1162,3 +1171,56 @@ message PayloadRecommendedSdkRemoved {
...
+  // issuer_url is the url of the OIDC issuer whose tokens the trust credential accepts.
+  string issuer_url = 2;
```
pkwarren: Should we include the `issuer` for well known issuers?

pkwarren (approved):

buf/registry/audit/v1/event.proto
```diff
@@ -1162,3 +1171,63 @@ message PayloadRecommendedSdkRemoved {
...
+message PayloadTrustCredentialCreated {
+  // user_id is the id of the user that owns the trust credential.
+  string user_id = 1 [(buf.validate.field).string.tuuid = true];
+  // issuer is the workload identity issuer whose OIDC tokens the trust credential accepts.
+  buf.registry.iam.v1.WorkloadIdentityIssuer issuer = 2;
+  // issuer_url is the url of the OIDC issuer whose tokens the trust credential accepts.
+  string issuer_url = 3;
+  // subject is the sub claim a workload's OIDC token must match, i.e. exactly which workloads this
+  // grant trusts.
+  string subject = 4;
```
nicksnyder, line 1184: should these have any protovalidate rules on them? (same questions for some other fields in this PR)

nicksnyder (approved):

buf/registry/audit/v1/event.proto
DMarby: We don't have protovalidate rules on any audit event fields outside of IDs currently (which these new events has also), feels fairly low value to add to other fields imo?

pkwarren: Yes - we don't need them as they are response only. I added the UUID rules more for documentation about the shape of the id strings.

nicksnyder: ok

